PR #31, opened 2018-11-18
Circleci

ali2251:


package.json
```diff
@@ -28,6 +28,7 @@
     "ganache-core": "^2.2.1",
     "request": "^2.88.0",
     "request-promise": "^4.2.2",
+    "truffle": "^4.1.14",
```
yahgwai: Lets add this to the state channel repo, then we can remove it from here.

.circleci/config.yml
```diff
@@ -0,0 +1,48 @@
...
+      - save_cache:
+          key: dependency-cache-{{ checksum "package.json" }}
+          paths:
+            - ./node_modules
+      - run:
+          name: build kitsune
+          command: npm run build-kitsune
+      - run:
+          name: test
+          command: npm run test
```
yahgwai: This currently doesnt include the test-service npm task, we should add that here as well or add it to the 'test' npm task

package.json
ali2251: sure

.circleci/config.yml
ali2251: if the test isnt doing anything interesting, then we can keep only one, either `test` or `service-test`

.circleci/config.yml
```diff
@@ -0,0 +1,48 @@
...
+      - run:
+          name: Deploy Over SSH
+          command: |
+            ssh $SSH_USER@$SSH_HOST "ls; cd mvp-api; git pull; node -v; yarn install; yarn build; yarn start"
+workflows:
+  version: 2
+  build-and-deploy:
+    jobs:
+      - build
+      - deploy:
```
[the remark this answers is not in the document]
ali2251: yup, will remove it

.circleci/config.yml
yahgwai: both are useful atm. We should run the functionality of both ```test``` and ```service-test```, either by combining them both into a command - or by just running both commands

ali2251: added both

.circleci/config.yml
```diff
@@ -0,0 +1,48 @@
...
+          key: dependency-cache-{{ checksum "package.json" }}
+          paths:
+            - ./node_modules
+      - run:
+          name: build kitsune
+          command: npm run build-kitsune
+      - run:
+          name: test
+          command: npm run test
+  deploy:
```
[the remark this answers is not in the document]
ali2251: we have both test and service test now

.circleci/config.yml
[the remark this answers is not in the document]
ali2251: done

ali2251:
Let me know if there is anything else

PR #40, opened 2018-12-10
Fix mocha

ali2251:
Major update to the structure of the repo and improved testing


ali2251:
should squash and merge!

yahgwai:
This looks great. I'm ready to merge, but why does it update files in the statechannels repo?

ali2251:
@yahgwai it moved state channel repo out to root of the directory, so git shows that as an add/delete

yahgwai:
But they shouldnt be committed right? The submodule should stop that from happening. We should just have the root folder committed I think.

Issue #29, opened 2018-11-15
Move implementation closer to paper spec


yahgwai:
The ts implementation should mirror as closely as possible the paper spec. This will aid in code review, auditing and general clarity. The focus in these code blocks should be on the security, and surrounding classes should focus on behaviour/UX.

yahgwai:
Fixed #45

PR #45, opened 2018-12-14
Separated the verifyAppointment critical section

yahgwai:
It looks like the verifyAppointment section is the only part of the paper we've implemented so far


yahgwai:
Fixed #29

src/inspector.ts
```diff
@@ -34,47 +36,28 @@ export class Inspector {
         logger.debug("Appointment request: " + JSON.stringify(appointmentRequest));
 
         const code: string = await this.provider.getCode(contractAddress);
+        // check that the channel is a contract
         if (code === "0x" || code === "0x00") {
-            throw new PublicInspectionError(`No code found at address: ${contractAddress}`);
-        }
-        if (code != this.deployedBytecode) {
```
ali2251: why is this check removed?

yahgwai:
Put the check back, tests now passing.

Issue #48, opened 2019-01-02
Docker image uses dev command


yahgwai:
The docker image currently uses the "start-dev" command when it should just use the "start" command. However using just the "start" command appears to cause a networking issue, which I havent been able to resolve.

yahgwai:
Fixed as part of #51

PR #49, opened 2019-01-02
Added docker image

yahgwai:


docker/docker_commands.txt
```diff
@@ -0,0 +1,22 @@
+// pisa net
```
ali2251: you can convert this a bash file and have commands as aliases exported to environment so they can be used more conveniently

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+COPY ./src ./src
+COPY ./tsconfig.json ./tsconfig.json
+# build
+RUN ["npm", "run", "build"]
+
+# expose the startup port
+EXPOSE 3000
+
+# start the application
+CMD ["npm", "run", "start-dev"]
```
ali2251: docker container should start production build instead of dev

config.json
```diff
@@ -1,7 +1,7 @@
 {
-    "jsonRpcUrl": "http://localhost:8545",
+    "jsonRpcUrl": "http://ganache:8545",
     "host": {
-        "name": "localhost",
+        "name": "0.0.0.0",
```
ali2251: this will expose your private ip address and anyone can then make a call to your blockchain

docker/docker-compose.yml
```diff
@@ -0,0 +1,24 @@
+version: "3"
+services:
+  ganache:
+    image: trufflesuite/ganache-cli:latest
+    ports:
+      - "8545:8545"
+    entrypoint: "node /app/ganache-core.docker.cli.js --mnemonic 'myth like bonus scare over problem client lizard pioneer submit female collect' -l 15000000 --allowUnlimitedContractSize --defaultBalanceEther 1000000"
```
ali2251: ideally, we dont want to be distributing mnemonics, eventhough they dont matter right now, this can be a problem in the future, i know we need it for testing but production should create a random one

package.json
```diff
@@ -4,11 +4,13 @@
...
-    "test": "npx tsc && mocha build/__pisa_tests__ --exit",
+    "test": "npm run build && mocha $(find ./build/__pisa_tests__/*.js ! -path '*quickTest*') --exit",
+    "test-docker": "mocha ./build/__pisa_tests__/quickTest.js --exit",
     "build-kitsune": "cd statechannels && npm ci && npx truffle compile",
     "start-evm": "ganache-cli -d -l 15000000 --allowUnlimitedContractSize --defaultBalanceEther 1000000",
-    "start": "ts-node src/ts/service.ts",
-    "build": "npx tsc",
+    "start-dev": "ts-node src/startUp.ts",
+    "start": "node ./build/startUp.js",
+    "build": "mkdir -p ./build && cp ./config.json ./build/ && tsc",
```
ali2251: instead of copying config file, its better to move the config file out of the ts folder to the root folder and reference it with `../` then all references remain intact even after compilation

package.json
```diff
@@ -4,11 +4,13 @@
   "description": "Arbitration Outsourcing for State Channels",
   "main": "index.js",
   "scripts": {
-    "test": "npx tsc && mocha build/__pisa_tests__ --exit",
+    "test": "npm run build && mocha $(find ./build/__pisa_tests__/*.js ! -path '*quickTest*') --exit",
```
ali2251: I dont understand why this needs to be done at all? wouldnt the command as it was work? it was execeuting all tests within __pisa_tests__ directory

package.json
```diff
@@ -4,11 +4,13 @@
...
+    "test-docker": "mocha ./build/__pisa_tests__/quickTest.js --exit",
```
ali2251: please dont use ./ when referring to a folder/file which is a sibling

src/__pisa_tests__/quickTest.ts
```diff
@@ -0,0 +1,90 @@
+import * as chai from "chai";
+import "mocha";
+import request from "request-promise";
+import { KitsuneTools } from "../kitsuneTools";
+import { ethers } from "ethers";
+import { IConfig } from "../dataEntities/config";
+const config = require("./../../config.json") as IConfig;
+const provider = new ethers.providers.JsonRpcProvider(config.jsonRpcUrl);
```
ali2251: please use the providers file to get the provider

src/provider.ts
```diff
@@ -1,7 +1,7 @@
 import Ganache from "ganache-core";
 import { ethers } from "ethers";
 import { IConfig } from "./dataEntities/config";
-const config = require("./config.json") as IConfig;
+const config = require("./../config.json") as IConfig;
```
ali2251: also when requiring a file in upper directory, just use `../`, its just more clearer

src/startUp.ts
```diff
@@ -3,11 +3,10 @@ import { ethers } from "ethers";
 import { IConfig } from "./dataEntities/config";
 import { KitsuneWatcher } from "./watcher";
 import { KitsuneInspector } from "./inspector";
-const config = require("./config.json") as IConfig;
+const config = require("./../config.json") as IConfig;
```
ali2251: same thing here `../` please

docker/docker_commands.txt
ali2251: Not necessary, this is just a comment

yahgwai: nice idea, I dont think I'll do it for now - but if I find myself using these commands a lot then I will

Dockerfile
yahgwai: agreed. I couldnt get the normal command to work though, it had been holding me up for half a day so I decided to put this in a raise a ticket for the change. See #48

config.json
yahgwai: That host refers to the PISA service, which we map to the ip of the container so that we can access it from outside the container. The docker container for ganache does this same thing though, so we'll have to check that we're not exposing ganache publicly - if we are it should be firewalled. This tutorial does a similar thing.

docker/docker-compose.yml
yahgwai: I think it's fine to use this mnemonic. Production shouldn't use Ganache - they should use a node, or a connection to infura

package.json
yahgwai: agreed. I made another ticket about #50 , feel free to add any thoughts to that

package.json
yahgwai: yeah, I didn't want it to execute the quick tests. These wont work when executed on circleci. I could just move them to another folder within source?

package.json
yahgwai: kk will do

src/provider.ts
yahgwai: kk will do

src/startUp.ts
yahgwai: kk will do

src/__pisa_tests__/quickTest.ts
yahgwai: agreed, added

Dockerfile
ali2251: cool

config.json
ali2251: cool

package.json
ali2251: yes please, it will be simpler that way

package.json
```diff
@@ -4,12 +4,13 @@
   "description": "Arbitration Outsourcing for State Channels",
   "main": "index.js",
   "scripts": {
-    "test": "npx tsc && mocha build/__pisa_tests__ --exit",
+    "test": "npm run build && mocha $(find build/__pisa_tests__/*.js ! -path '*quickTest*') --exit",
```
ali2251: this can be changed back to the way it was now

ali2251: since we have test-docker for testing docker

ali2251:
looks all good, just one tiny package.json change then ready to be merged

package.json
yahgwai: cool, done

package.json
yahgwai: done

yahgwai:
Ready for merge when it passes review

Issue #47, opened 2018-12-17
Create a deployment package


yahgwai:
Probably using docker. Create a package that can be deployed onto any machine to host PISA

ali2251:
makes sense to have a docker image that can be started by anyone

PR #53, opened 2019-01-03
51 staged docker

yahgwai:


src/__docker_tests__/quickTest.ts
```diff
@@ -84,8 +75,13 @@ const waitForPredicate = <T1>(successResult: T1, predicate: (a: T1) => boolean,
...
+                reject("Predicate not satisfied.");
             }
         }, timeout);
     });
 };
+
+console.log("Executing smoke tests");
+setup().then(() => {
+    execute().then(() => console.log("Tests passed!"));
+});
```
ali2251: this approach to test works, but i dont think its scalable, is there any particular reason for not using mocha? or is it just an optimisation

src/inspector.ts
```diff
@@ -34,8 +34,9 @@ export class Inspector {
             `Inspecting appointment ${appointmentRequest.stateUpdate.hashState} for contract ${contractAddress}.`
         );
         logger.debug("Appointment request: " + JSON.stringify(appointmentRequest));
-        
+        console.log("hello")
```
ali2251: can be removed

src/inspector.ts
```diff
@@ -34,8 +34,9 @@ export class Inspector {
...
         const code: string = await this.provider.getCode(contractAddress);
+        console.log("goodbye")
```
ali2251: can be removed (console.log)

src/logger.ts
```diff
@@ -13,9 +13,9 @@ const logger = createLogger({
     level: "info",
     format: combinedFormats,
     transports: [
```
ali2251: I would say that logger *must* handle creating the logs directory itself and not rely on the script running it create it

src/logger.ts
```diff
@@ -13,9 +13,9 @@ const logger = createLogger({
...
-        new transports.File({ filename: "error.log", level: "error" }),
-        new transports.File({ filename: "info.log", level: "info" }),
-        new transports.File({ filename: "debug.log", level: "debug" })
+        new transports.File({ dirname: "logs", filename: "error.log", level: "error" }),
+        new transports.File({ dirname: "logs", filename: "info.log", level: "info" }),
+        new transports.File({ dirname: "logs", filename: "debug.log", level: "debug" })
```
ali2251: I am not sure if there is a benefit of having 3 different files, having all logs in a single file often makes debugging a lot easier

src/__docker_tests__/quickTest.ts
yahgwai: I agree. I removed mocha temporarily to reduce the dependencies required to run tests - this was initially because tests were part of the deployment. I've created a new 'test' docker image to contain these test files, and it will contain the relevant environment - mocha - to run integration tests. So we'll migrate back to mocha on a later branch.

src/inspector.ts
yahgwai: removed

src/inspector.ts
yahgwai: removed

src/logger.ts
yahgwai: agreed. Will do that here.

src/logger.ts
yahgwai: Hmm, I take your point. The debug log does contain all the outputs though - we're logging more than we need to. I think it's quick nice to see just the errors, or just the info sometimes though.

ali2251:
Why is the base branch of this PR and other one is not master?

yahgwai:
I based off of this branch as it had dependencies on the work I'd done in this branch. So my branches had this dependency, so I mirrored it in the PRs. If I had based off of master you would have seen the the diff between master and #47 unioned with the diff between #47 and this, making it hard to review. Unless you merged #47 before starting this review.

We should have a chat about this process, if there's a better way I'm happy to adopt it.

PR #57, opened 2019-01-04
Re-organised some files, and added a very basic integration test

yahgwai:
This just a first test, and a stab at a workflow on how we could add more.


configs/test.json
```diff
@@ -0,0 +1,19 @@
+{
```
ali2251: Its a good idea to think about whether we want to have multiple config files or a single config file with all configs of all environments

src/kitsuneTools.ts
```diff
@@ -17,12 +26,37 @@ export class KitsuneTools {
...
+        try {
+            logger.debug("a1");
+
+            let trying = true;
+            let tries = 0;
+            let tx;
+            while (trying && tries < 10) {
+                try {
+                    tx = await contract.setstate(
+                        [sig0.v - 27, sig0.r, sig0.s, sig1.v - 27, sig1.r, sig1.s],
```
ali2251: I thought we add 27 to v .....

src/kitsuneTools.ts
```diff
@@ -17,12 +26,37 @@ export class KitsuneTools {
...
+                        appointment.stateUpdate.round,
+                        appointment.stateUpdate.hashState
+                    );
+                    trying = false;
+                } catch (exe) {
+                    // lets retry this hard until we can no longer
```
ali2251: you may want to log the actual error too

src/kitsuneTools.ts
```diff
@@ -17,12 +26,37 @@ export class KitsuneTools {
...
+                    logger.error(`Failed to set state for contract ${contract.address}, re-tries ${tries}`)
+                    tries++;
+                    await wait(1000);
+                }
+            }
+
+            if(trying) throw new Error("Failed after 10 tries.")
```
ali2251: why not do if (tries > 10) instead of using a separate environment variable

src/customExpressHttpContext.ts
```diff
@@ -3,7 +3,7 @@ import uuid from "uuid";
 
 const REQUEST_ID = "requestId";
 
-export const setRequestId = () => {
+export const setRequestId = (value? : string) => {
     return httpContext.set(REQUEST_ID, uuid.v1());
```
ali2251: the reason it maybe isnt working is because you are calling v1 function, i havent tried it so its just a wild guess, but on a side note, i suggest to use v4 (random) because v1 is predictable

test/load/loadTest.ts
```diff
@@ -0,0 +1,144 @@
+import request from "request-promise";
+import { KitsuneTools } from "../../src/kitsuneTools";
+import { ethers } from "ethers";
+import { IConfig } from "../../src/dataEntities/config";
+import { getJsonRPCProvider } from "../../src/provider";
+let config = require("../../config.json") as IConfig;
+// let config = {
```
ali2251: the commented code can be removed if not used

test/load/loadTest.ts
```diff
@@ -0,0 +1,144 @@
...
+let account0: string,
+    account1: string,
+    channelContracts: ethers.Contract[] = [],
+    hashState: string,
+    disputePeriod: number;
+
+// concurrently watch for x events
+let setup = async (creationCount: number) => {
+    const provider = await getJsonRPCProvider();
+    if (creationCount === 0) {
```
ali2251: you can change the condition to <1 so all negative numbers are accounted for as well

configs/test.json
yahgwai: I'm thinking of making another repo - it would contain: 
1) The docker compose files
2) The config files
Then pisa would stay as pisa, and this other repo would be a kind of 'deployment' repo. What dyou think?

A different way would be to create a *package* step. This would build the docker image, publish it, then create a package that contains the docker compose files and an example config.

src/kitsuneTools.ts
yahgwai: The contract expects 0 or 1. It then adds 27.

splitsig creates 27 or 28

src/customExpressHttpContext.ts
yahgwai: hmm, I'll check that

src/kitsuneTools.ts
yahgwai: good catch - this whole section of code shouldn't have been committed :s

We should be doing retry logic in the watcher, not in the kitsune tools.

Interestingly I had to do this to even get the basic load test to pass. Many events require multiple retries when they're happening concurrently - I think the culprit is ganache.

configs/test.json
ali2251: I prefer the package creation instead of another repository, the easiest way to do this is to save the packages in circleci artifacts and a link can be shared for anyone willing to download

yahgwai: OK, agreed. I'll try to organise the configs for this.

PR #59, opened 2019-01-06
Clean package.json

ali2251:


package.json
```diff
@@ -10,7 +10,7 @@
     "start-evm": "ganache-cli -d -l 15000000 --allowUnlimitedContractSize --defaultBalanceEther 1000000",
     "dev": "ts-node src/startUp.ts",
     "start": "node ./build/startUp.js",
-    "build": "mkdir -p ./build && cp ./config.json ./build/ && tsc",
+    "build": "tsc",
```
yahgwai: When I run "npm run build" now, the result is a build directory without a config.json in it? Is this expected? If I also run npm run start after I get an error due to this missing conifg.

Btw, I tried to get the tsconfig option resolveJsonModule to work, but had no luck so I left this here.

ali2251: i tried it and wasnt getting any error, all of the config requires are "../" which means it should be fine to use from the build folder as all thats happening is src/ is converted to build.

ali2251: If you want to build a dist/ package for docker, then i suggest to do it separately, since paths would then have to care about other things too, i will double check the issue sometime today

PR #60, opened 2019-01-06
Revert "51 staged docker"

ali2251:
Reverts PISAresearch/pisa#53


ali2251:
if the behaviour you wanted was actually to merge with other branches rather than master, then you can ignore the reverts and make a single PR to merge the master, I would need some of the changes for the things i am working on

PR #101, opened 2019-04-08
Added unit tests for the raiden inspector

yahgwai:


src/integrations/raiden/index.ts
```diff
@@ -167,13 +167,13 @@ export class RaidenInspector extends Inspector<RaidenAppointment> {
 
         const code: string = await this.provider.getCode(contractAddress);
         // check that the channel is a contract
-        if (code === "0x" || code === "0x00") {
+        if (!code || code === "0x" || code === "0x00") {
             throw new PublicInspectionError(`No code found at address ${contractAddress}`);
         }
```
bigspider: Is the behavior of returning "0x" or "0x00" observed, or is it a bug in ethers.js? I don't see it documented there.

test/mocked/raidenInspector.test.ts
```diff
@@ -0,0 +1,236 @@
...
+        // contract
+        const tokenNetworkAddress = "0x34636d2289588BcA4e0fcF863857386F56b44bdC";
+        const party1Address = "0xccca21b97b27DefC210f01A7e64119A784424D26";
+        const party2Address = "0xdddEC4D561eE68F37855fa3245Cb878b10Eb1fA0";
+        const otherTokenNetwork = "0xf0afbed24d88ce4cb12828984bb10d2f1ad0e185"
+        const mockedProvider = mockito.mock(ethers.providers.JsonRpcProvider);
+        mockito.when(mockedProvider.getCode(tokenNetworkAddress)).thenResolve(RaidenTools.ContractDeployedBytecode);
+        mockito.when(mockedProvider.getCode(otherTokenNetwork)).thenResolve(RaidenTools.ContractDeployedBytecode);
+        mockito.when(mockedProvider.getCode(party1Address)).thenResolve("0x");
+        mockito.when(mockedProvider.getCode(party2Address)).thenResolve(RaidenTools.ContractDeployedBytecode.slice(0, RaidenTools.ContractDeployedBytecode.length - 20));
```
bigspider: Add a comment to note that getCode is mocking incorrect behavior for party1Address and party2Address (easy to miss)

test/mocked/raidenInspector.test.ts
```diff
@@ -0,0 +1,236 @@
...
+
+        const appointment = new RaidenAppointment({
+            ...appointmentObj,
+            stateUpdate: { ...appointmentObj.stateUpdate, non_closing_signature: incorrectNonClosingParticipantSig }
+        });
+
+        const inspector = new RaidenInspector(minimumDisputePeriod, provider);
+        await isRejected(inspector.checkInspection(appointment));
+    });
+
```
bigspider: Should we add a couple of tests that checksummed addresses are accepted in the appointment? (More general question: do we want to force users to provide checksummed addresses in our API?)

src/integrations/raiden/index.ts
yahgwai: I'm only observing 0x, I'll update

test/mocked/raidenInspector.test.ts
yahgwai: We should add those tests, but currently that code exists in the appointment itself. We need to write a bunch of separate tests for the parsing code that exists there.

I would suggest that we probably should force the check sum, the more constrained the data is the better generally.

test/mocked/raidenInspector.test.ts
```diff
@@ -0,0 +1,236 @@
...
+const isRejected = async (result: Promise<any>) => {
+    return await result.then(
+        () => {
+            assert.fail();
+        },
+        reject => {
+            expect(reject).to.exist;
+        }
+    );
+};
```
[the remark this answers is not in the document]
yahgwai: Nice! Updated and using that now.

test/mocked/raidenInspector.test.ts
yahgwai: Done

PR #107, opened 2019-04-12
102 watcher refactor tests

yahgwai:


src/service.ts
```diff
@@ -1,16 +1,17 @@
 import express, { Response } from "express";
 import httpContext from "express-http-context";
 import logger from "./logger";
-import { PublicInspectionError, PublicDataValidationError, IEthereumResponse } from "./dataEntities";
+import { PublicInspectionError, PublicDataValidationError, ApplicationError } from "./dataEntities";
 import { Raiden, Kitsune } from "./integrations";
-import { Watcher } from "./watcher";
+import { Watcher } from "./watcher/watcher";
```
bigspider: Perhaps I would add a ./watcher/index.ts to simplify imports.

src/watcher/appointmentSubscriber.ts
```diff
@@ -0,0 +1,96 @@
...
+    (...args: any[]): void;
+    appointmentId: string;
+}
+
+/**
+ * If we wish to unsubscribe an appointment from the provider we have a problem. In the provider
+ * appointments are keyed only by their filter, so an appointment could be updated, then another
+ * process tries to remove the old appointment and in doing so removes the new one. This class
+ * assigns appointment ids to listeners and ensures that appointments are only removed from the
+ * provided if the provide appointment id matches.
```
bigspider: Typo in the last sentence of the comment.

src/watcher/garbageCollector.ts
```diff
@@ -0,0 +1,78 @@
...
+                        logger.info(a.formatLog(`GC: Appointment with end: ${a.endTime}.`))
+                    );
+                }
+
+                this.collecting = false;
+            }
+        } catch (doh) {
+            // errors escaping this block will cause unhandled promise rejections so we log and swallow here
+            logger.error("GC: Unexpected error.");
+            logger.error(doh);
```
bigspider: I think the lock should released outside of the try/catch, or it will be permanently locked in case of error.

yahgwai: good catch

src/service.ts
yahgwai: will do

src/watcher/store.ts
```diff
@@ -0,0 +1,90 @@
...
+                delete this.appointmentsByStateLocator[appointmentById.getStateLocator()];
+            }
+            return true;
+        }
+        return false;
+    }
+
+    async getExpiredSince(expiryTime: number): Promise<IEthereumAppointment[]> {
+        return Object.keys(this.appointmentsById)
+            .map(a => this.appointmentsById[a])
```
bigspider: Aren't the last two lines equivalent to ```Object.values(this.appointmentsById)```?

src/watcher/appointmentSubscriber.ts
yahgwai: fixed

src/watcher/store.ts
yahgwai: Yeah. It seems we needed to switch tscongif to use the es2017 before we could use. Done that now.

PR #110, opened 2019-04-18
Tests to go with the recent watcher refactoring

yahgwai:


src/watcher/garbageCollector.ts
```diff
@@ -23,52 +24,74 @@ export class AppointmentStoreGarbageCollector {
...
-    private collecting = false;
+    /**
+     * Stop monitoring for expired events
+     */
+    public stop() {
+        if (this.started) {
+            this.started = false;
+            this.provider.removeListener("block", this.listener);
+        }
+    }
```
bigspider: Should we throw an error if start is called when it's already started, or stop when not started? Isn't it likely a bug if that happens?

src/watcher/garbageCollector.ts
```diff
@@ -23,52 +24,74 @@ export class AppointmentStoreGarbageCollector {
...
 
     /**
      * Find appointments that have expired then remove them from the subsciber and store.
      * @param blockNumber
```
bigspider: typo: subscriber

src/watcher/garbageCollector.ts
yahgwai: cool, I'll update

src/watcher/garbageCollector.ts
yahgwai: updated

src/watcher/garbageCollector.ts
yahgwai: I'll throw error for "start" but not for "stop" as errors in destructors can be dangerous. But I'll log error there.

PR #126, opened 2019-05-02
122 rate limit

bigspider:
Added new optional key config.endPoint to add any endpoint settings. The only two settings (both optional) are currently rateGlobal and ratePerUser; each of them, has compulsory "msWindow" and "max" and optional "message". Per user rate limit return code 429 when hit, while the global limits return 503.

In the default config.json, I only added a default per user limit of 5 requests per IP per minute and no global rate limit.

If a PisaTower is created with no 'config' parameter, rate limits are disabled by default (but a warning is logged in that case).


src/dataEntities/config.ts
```diff
@@ -1,12 +1,24 @@
 import config from "../config.json";
 
+interface IRateConfig {
+    windowMs: number;
+    max: number;
+    message?: string;
```
yahgwai: This isn't used. And I agree that it shouldn't be, I don't think we need to allow configurability of the error message.

src/service.ts
```diff
@@ -47,6 +54,30 @@ export class PisaService {
...
+                ...config.rateGlobal
+            }));
+            logger.info(`PISA api global rate limit: ${config.rateGlobal.max} requests every: ${config.rateGlobal.windowMs/1000} seconds.`);
+        } else {
+            logger.warn(`PISA api global rate limit: NOT SET.`);
+        }
+
+        if (config && config.ratePerUser) {
+            app.use(new rateLimit({
+                statusCode: 429, // = Too Many Requests (RFC 6585)
```
yahgwai: I know the default is already to IP, but it would be easier to read if we put the keyGenerator value in: keyGenerator: (req) => req.ip.

src/service.ts
```diff
@@ -47,6 +54,30 @@ export class PisaService {
             next();
         });
 
+        // rate limits
+        if (config && config.rateGlobal) {
```
yahgwai: The middleware section is quite large now, lets move all app.use() methods into a new function applyMiddleware(app) just so that the constructor is a bit easier to read.

src/service.ts
```diff
@@ -1,5 +1,8 @@
 import express, { Response } from "express";
 import httpContext from "express-http-context";
+
```
yahgwai: You don't need this whitespace

src/config.json
```diff
@@ -4,5 +4,12 @@
         "name": "localhost",
         "port": 3000
     },
-    "responderKey": "0x6370fd033278c143179d81c5526140625662b8daa446c22ee2d73db3707e620c"
+    "responderKey": "0x6370fd033278c143179d81c5526140625662b8daa446c22ee2d73db3707e620c",
+    "apiEndpoint": {
```
yahgwai: Lets set a default global rate here as well. I think we should also add these options to the command line args, just so that we have parity between them.

src/dataEntities/config.ts
bigspider: According to standards, status 429 should include details about what rate limit is in place (in fact I put a custom message in the default config.json.

src/service.ts
bigspider: Done

src/service.ts
bigspider: Done.

src/service.ts
bigspider: Done; while I was there, I reordered imports, so that all libraries are before local imports.

src/service.ts
```diff
@@ -47,6 +54,30 @@ export class PisaService {
             next();
         });
 
+        // rate limits
+        if (config && config.rateGlobal) {
+            app.use(new rateLimit({
+                keyGenerator: () => "global", // use the same key for all users
+                statusCode: 503, // = Too Many Requests (RFC 7231)
+                message: "PISA is overloaded right now. Please try again later.",
```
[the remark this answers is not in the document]
bigspider: Done. Also changed the message for user's rate limit in the default config.json, which was even more informal.

src/config.json
bigspider: Done, set 100 request per minute in the default config.json.

PR #134, opened 2019-05-06
Enable more restrictive compiler options

bigspider:
Enabled options noImplicitAny and strictNullChecks, and updated codebase accordingly.
Fixed a couple of potential bugs in the process.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
   "devDependencies": {
     "@types/chai": "^4.1.6",
     "@types/chai-as-promised": "^7.1.0",
+    "@types/cls-hooked": "^4.3.0",
     "@types/express": "^4.16.0",
     "@types/express-rate-limit": "^3.3.0",
     "@types/fs-extra": "^5.0.5",
--- a/src/blockMonitor/reorg.ts
+++ b/src/blockMonitor/reorg.ts
@@ -1,5 +1,5 @@
 import { ethers } from "ethers";
-import { StartStopService } from "../dataEntities";
+import { StartStopService, ApplicationError } from "../dataEntities";
 import logger from "../logger";
 import { BlockStubChain, IBlockStub } from "./blockStub";
 import { ReorgHeightListenerStore } from "./reorgHeightListener";
@@ -53,7 +53,7 @@ export class ReorgDetector extends StartStopService {
      * Detect a reorg if a new block is observed
      * @param blockNumber
      */
-    private async handleNewBlock(blockNumber) {
+    private async handleNewBlock(blockNumber: number) {
         // we should lock here so that we dont fire reorg events concurrently
         // it doesnt matter if a reorg is missed immediately, it will be picked up on the next block emission
         if (!this.handlingBlock) {
@@ -97,7 +97,7 @@ export class ReorgDetector extends StartStopService {
                 logger.error(`${this.name}: Unexpected error.`);
                 const dohError = doh as Error;
                 if (dohError) {
-                    logger.error(dohError.stack);
+                    logger.error(dohError.stack!);
                 }
             } finally {
                 this.handlingBlock = false;
@@ -180,7 +180,7 @@ export class ReorgDetector extends StartStopService {
         localBlock: BlockStubChain,
         differenceBlocks: IBlockStub[],
         minHeight: number
-    ): Promise<BlockStubChain> {
+    ): Promise<BlockStubChain | null> {
         const blockRemote = await this.provider.getBlock(remoteBlockHash);
         differenceBlocks.push(blockRemote);
 
@@ -201,8 +201,12 @@ export class ReorgDetector extends StartStopService {
     public async findCommonAncestor(
         newBlock: IBlockStub,
         currentHead: BlockStubChain
-    ): Promise<{ commonAncestor: BlockStubChain; differenceBlocks: IBlockStub[] }> {
-        let commonAncestor: BlockStubChain;
+    ): Promise<{ commonAncestor: BlockStubChain | null; differenceBlocks: IBlockStub[] }> {
+        if (newBlock.parentHash === null) {
+            throw new ApplicationError("newBlock should have a parentHash");
+        }
+
+        let commonAncestor: BlockStubChain | null;
         let differenceBlocks: IBlockStub[] = [];
         const minHeight = this.headBlock.height - this.maxDepth;
         // the chain has reduced linearly
--- a/src/blockMonitor/reorg.ts
+++ b/src/blockMonitor/reorg.ts
@@ -1,5 +1,5 @@
 import { ethers } from "ethers";
-import { StartStopService, ApplicationError } from "../dataEntities";
+import { StartStopService, ArgumentError } from "../dataEntities";
 import logger from "../logger";
 import { BlockStubChain, IBlockStub } from "./blockStub";
 import { ReorgHeightListenerStore } from "./reorgHeightListener";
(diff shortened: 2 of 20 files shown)


src/dataEntities/appointment.ts
```diff
@@ -60,11 +61,15 @@ export abstract class EthereumAppointment implements IEthereumAppointment {
...
         this.mPassedInspection = passed;
         if (passed) {
-            this.mStartBlock = startBlock;
-            this.mEndBlock = startBlock + this.expiryPeriod;
+            if (startBlock === null) {
+                throw new ApplicationError("startBlock must be provided on a passed inspection.");
+            } else {
+                this.mStartBlock = startBlock;
+                this.mEndBlock = startBlock + this.expiryPeriod;
+            }
```
bigspider: Here I'm assuming that startBlock is ignored if `passed === false`, but it must not be `null` otherwise.

src/inspector.ts
```diff
@@ -5,14 +5,14 @@ import logger from "./logger";
 export abstract class Inspector<TAppointment extends EthereumAppointment> {
     protected constructor(public readonly channelType: ChannelType, protected readonly provider: ethers.providers.Provider) {}
 
-    public abstract async checkInspection(appointment): Promise<void>;
+    public abstract async checkInspection(appointment: TAppointment): Promise<void>;
```
bigspider: Here (and below), the parametric type TAppointment was not used.

src/responder.ts
```diff
@@ -87,8 +87,16 @@ export abstract class EthereumResponder extends Responder {
...
+    protected provider: ethers.providers.Provider;
+
     constructor(public readonly signer: ethers.Signer) {
         super();
+
+        if (!signer.provider) {
+            throw new ApplicationError("The given signer is not connected to a provider");
+        } else {
+            this.provider = signer.provider;
+        }
```
bigspider: I was always assuming that the provider is available on a Signer, but it's not the case; thus, now I'm checking in the constructor.

src/tower.ts
```diff
@@ -17,7 +17,7 @@ export class PisaTower {
     }
 
     configs: {
-        [type: number]: IChannelConfig<EthereumAppointment, Inspector<EthereumAppointment>>;
+        [type: string]: IChannelConfig<EthereumAppointment, Inspector<EthereumAppointment>>;
```
bigspider: The type was incorrectly declared as `number` rather than the `string`.

test/integration/raiden.test.ts
```diff
@@ -4,7 +4,7 @@ import * as path from "path";
 import * as fse from "fs-extra";
 import net from "net";
 
-import waitPort from "wait-port";
+const waitPort: any = require("wait-port");
```
bigspider: wait-port does not have proper type declarations; using _require_ allows to disable the type checking.

src/dataEntities/appointment.ts
yahgwai: This method seems a bit outdated. It seems we never used the passed: false variant. So lets refactor this method, call it passInspection(startBlock: number), and not allow startBlock to be null.

yahgwai: As a side note, we have an ArgumentError error that you can throw when checking arguments.

src/responder.ts
yahgwai: Seems sensible. You could use ArgumentError here. Also, if you want to, you can make this a little more concise by not using the else here:

   if (!signer.provider) throw new ApplicationError("The given signer is not connected to a provider");
   this.provider = signer.provider;

yahgwai: It also seems that EthereumResponder no longer requires the provider - I think this is since you refactored the submitStateFunction to be prepareRequest.

package.json
```diff
@@ -48,6 +48,7 @@
   "devDependencies": {
     "@types/chai": "^4.1.6",
     "@types/chai-as-promised": "^7.1.0",
+    "@types/cls-hooked": "^4.3.0",
```
yahgwai: What's this used for?

src/blockMonitor/reorg.ts
```diff
@@ -201,8 +201,12 @@ export class ReorgDetector extends StartStopService {
     public async findCommonAncestor(
         newBlock: IBlockStub,
         currentHead: BlockStubChain
-    ): Promise<{ commonAncestor: BlockStubChain; differenceBlocks: IBlockStub[] }> {
-        let commonAncestor: BlockStubChain;
+    ): Promise<{ commonAncestor: BlockStubChain | null; differenceBlocks: IBlockStub[] }> {
+        if (newBlock.parentHash === null) {
```
yahgwai: ArgumentError. I think you could even extend ArgumentError with a new error, ArgumentNullError and use that when checking for null.

src/responder.ts
```diff
@@ -443,13 +465,13 @@ export class EthereumResponderManager {
                 // Should we keep inactive responders anywhere?
                 this.responders.delete(responder);
             })
-            .on(ResponderEvent.AttemptFailed, (responseFlow: ResponseFlow, doh, attemptNumber) => {
+            .on(ResponderEvent.AttemptFailed, (responseFlow: ResponseFlow, doh, attemptNumber: number) => {
```
yahgwai: doh doesnt fail here because on() accepts ...args: any[]? We should probably insist that it's of type error anyway, since it would be a bug if we were throwing an object that wasnt.

src/tower.ts
yahgwai: Good catch!

test/src/blockMonitor/reorg.test.ts
```diff
@@ -123,9 +123,9 @@ class TestCase {
         );
 
     async traverse(reorgDetector: ReorgDetector, provider: asyncEmitTestProvider) {
-        const findReorgSpec = blockNumber =>
+        const findReorgSpec = (blockNumber: number) =>
             this.reorgs.filter(r => r.expectedAtBlockNumber === blockNumber && !r.observed)[0];
-        let currentReorg;
+        let currentReorg: any;
```
yahgwai: I think this should be IReorgInfo

test/src/responder.test.ts
```diff
@@ -94,9 +102,9 @@ async function getTestData(provider, account0, account1, responderAccount, hashS
 const _setTimeout = global.setTimeout;
 
 // Save a snapshot of the state of the blockchain in ganache; resolves to the id of the snapshot
-function takeGanacheSnapshot(ganache): Promise<string> {
+function takeGanacheSnapshot(ganache: any): Promise<string> {
```
yahgwai: It seems Ganache has really bad typings :(

package.json
bigspider: I get this error without it, didn't investigate further:
```
node_modules/express-http-context/index.d.ts:2:27 - error TS7016: Could not find a declaration file for module 'cls-hooked'. '/home/spider/prog/pisa/node_modules/cls-hooked/index.js' implicitly has an 'any' type.
  Try `npm install @types/cls-hooked` if it exists or add a new declaration (.d.ts) file containing `declare module 'cls-hooked';`

2 import { Namespace } from 'cls-hooked';
```

src/dataEntities/appointment.ts
bigspider: Refactored as suggested.

src/blockMonitor/reorg.ts
bigspider: Changed to ArgumentError.

test/src/blockMonitor/reorg.test.ts
bigspider: Changed to `IReorgInfo | undefined`, since it might be used before initializing it.

src/responder.ts
bigspider: Done.

EthereumResponder does not directly use the provider, but I expect any subclass to use it, so I'd leave it there as a protected member for now.

package.json
yahgwai: Fair enough

PR #138, opened 2019-05-09
Added a leveldb backend for appointments

yahgwai:


src/service.ts
```diff
@@ -41,28 +44,44 @@ export class PisaService {
...
+        const configs = [Raiden, Kitsune]
+
         // start reorg detector
         this.reorgDetector = new ReorgDetector(delayedProvider, 200, new ReorgHeightListenerStore());
-        this.reorgDetector.start();
 
         // dependencies
-        const store = new MemoryAppointmentStore();
+        const appointmentConstructors = new Map<ChannelType, (obj: any) => IEthereumAppointment>()
+        configs.map(c => appointmentConstructors.set(c.channelType, c.appointment));
```
bigspider: I think `for (let c of configs) { ... }` would be better here.

src/dataEntities/startStop.ts
```diff
@@ -38,10 +38,10 @@ export abstract class StartStopService extends EventEmitter {
     /**
```
bigspider: I agree start and stop sthould be async, but one change is needed:
- As it is async, nothing prevents different objects to call `start()` again before `startInternal()` completes, so that `mStarted` is still false. 

While this is unrelated to this commit, I am thinking of this other change:
- As it could make sense to share the same service in two other objects (e.g. two different objects might want to use the same ReorgDetector), it could make sense to change the semantic from start/stop to something like attach/detach. That is, the service actually stops when everyone detaches.

src/startUp.ts
```diff
@@ -103,11 +135,13 @@ function waitForStop(service: PisaService) {
...
         stdin.setEncoding("utf8");
-        stdin.on("data", key => {
+        stdin.on("data", async key => {
             // ctrl-c ( end of text )
             if (key === "\u0003") {
                 // stop the pisa service
-                service.stop();
+                await service.stop();
+                // shut the db
+                await db.close();
```
bigspider: You could increase parallelism:
```
await Promise.all([
  service.stop(),
  db.close()
]);
```
Although it's slightly less readable.

src/watcher/store.ts
```diff
@@ -22,23 +24,56 @@ export interface IAppointmentStore {
...
+        super("Appointment store");
+    }
+
+    async startInternal() {
+        // access the db and load all state
+        for await (const record of this.db.createValueStream()) {
+            // the typing here insist this is a string
+            const type = ((record as any) as IAppointment).type;
+            const constrctr = this.appointmentConstructors.get(type);
+            if (!constrctr) throw new ConfigurationError(`Unrecognied channel type: ${type}.`);
```
bigspider: typo, "Unrecognied"

src/watcher/watcher.ts
```diff
@@ -7,6 +7,12 @@ import { ethers } from "ethers";
 import { EthereumResponderManager } from "../responder";
 import { ReorgDetector } from "../blockMonitor";
 import { inspect } from "util";
+import levelup, { LevelUp } from "levelup";
+import leveldown from "leveldown";
+import encodingDown from "encoding-down";
+import EncodingDown from "encoding-down";
+import { Readable } from "stream";
+import { ReadStream, createReadStream } from "fs";
```
bigspider: All these imports are not used in this file.

src/service.ts
yahgwai: I also didn't like my initialisation. I've figured out how to use the other constructor. What do you think of this:

new Map(configs.map(c => [c.channelType, c.appointment]))

I think it's quite clear and concise

bigspider: Yep, looks good.

src/dataEntities/startStop.ts
yahgwai: Added an mStarting var, and a test to ensure it works.

src/startUp.ts
yahgwai: updated

src/watcher/store.ts
yahgwai: done

src/watcher/watcher.ts
yahgwai: removed

src/dataEntities/startStop.ts
```diff
@@ -18,19 +18,29 @@ export abstract class StartStopService extends EventEmitter {
...
     }
     private mStarted: boolean = false;
+    private mStarting: boolean = false;
 
     /**
      * Start this service
      */
     public async start() {
         if (this.mStarted) throw new ConfigurationError(`${this.name}: Already started.`);
+        if (this,this.mStarting) throw new ConfigurationError(`${this.name}: Currently starting.`);
```
bigspider: typo, remove `this,`

PR #142, opened 2019-05-09
Add appointment signatures to Pisa tower responses

bigspider:
Added receiptKey in config as a key to sign (Ethereum) accepted appointments.

PisaService uses a receiptKey in the config to create a Wallet.
PisaTower, instead, expects an EthereumAppointmentSigner, that knows how to sign appointments.

The reason for the different interface is that future PisaService implementations _will not_ have access to the signer at all, but it useful for now, since a pisa service bundles all the pieces together in one service. Instead, the Tower _will_ have access to some signing authority, but how to get the appointments signed is out of its concern and encapsulated in the EthereumAppointmentSigner.

### DIFF
--- a/src/service.ts
+++ b/src/service.ts
@@ -15,7 +15,7 @@ import {
 } from "./dataEntities";
 import { Raiden, Kitsune } from "./integrations";
 import { Watcher, AppointmentStore } from "./watcher";
-import { PisaTower } from "./tower";
+import { PisaTower, HotEthereumAppointmentSigner } from "./tower";
 import { setRequestId } from "./customExpressHttpContext";
 import { EthereumResponderManager } from "./responder";
 import { AppointmentStoreGarbageCollector } from "./watcher/garbageCollector";
@@ -42,6 +42,7 @@ export class PisaService extends StartStopService {
      * @param port The port on which to host the pisa service
      * @param provider A connection to ethereum
      * @param wallet A signing authority for submitting transactions
+     * @param receiptSigner A signing authority for receipts returned from Pisa
      * @param delayedProvider A connection to ethereum that is delayed by a number of confirmations
      * @param config Optional configuration of the Pisa endpoint
      */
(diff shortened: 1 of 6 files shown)


src/service.ts
```diff
@@ -168,11 +173,16 @@ export class PisaService extends StartStopService {
...
+                const { appointment, signature } = await tower.addAppointment(req.body);
 
                 // return the appointment
                 res.status(200);
-                res.send(appointment);
+
+                // with signature
+                res.send({
+                    appointment,
+                    signatures: [signature]
```
yahgwai: why plural signature and an array?

src/tower.ts
```diff
@@ -42,6 +43,52 @@ export class PisaTower {
...
+        super();
+    }
+
+    /**
+     * Signs `appointment`. Returns a promise that resolves to the signature of the appointment.
+     *
+     * @param appointment
+     */
+    public signAppointment(appointment: EthereumAppointment): Promise<string> {
+        // TODO: this signature format does not match Pisa's contract
```
yahgwai: Lets remove this TODO from the code, and add it as a ticket.

src/service.ts
bigspider: While we only have one now, aren't we going to give multiple signatures in the receipt?

yahgwai: Maybe there's some confusion coming from the fact that this class is called PisaService, when actually if we look at what it does perhaps a better name would be PisaTowerService. A pisa tower should probably only have one signing key, then we can aggregate those signatures as part of some other component.

bigspider: Currently, PisaService hosts the `/appointment` endpoint, which is the user/customer's interface, so I think it makes sense to leave its interface (and response) to what the user expects. When we build a (Pisa)TowerService, then PisaService can communicate with that instead. So I think the component that returns one signature should be the (yet missing) one, not the api endpoint.

src/tower.ts
bigspider: Done.

src/service.ts
bigspider: Done, put directly the signature without the array.

PR #150, opened 2019-05-13
Added a Lock util; synchronizing accesses by appointment's state locator in the AppointmentStore.

bigspider:


src/watcher/store.ts
```diff
@@ -116,16 +121,17 @@ export class AppointmentStore extends StartStopService implements IAppointmentSt
...
+            await this.lockManager.withLock(appointmentById.getStateLocator(), async () => {
+                // remove the appointment from the state locator index
+                const currentAppointment = this.appointmentsByStateLocator[appointmentById.getStateLocator()];
+                // if it has the same id
+                if (currentAppointment.id === appointmentId) {
+                    delete this.appointmentsByStateLocator[appointmentById.getStateLocator()];
+                }
+
+                // and remove from remote storage
+                await this.db.del(appointmentId);
```
yahgwai: I dont think this needs to be inside the lock

src/watcher/store.ts
```diff
@@ -116,16 +121,17 @@ export class AppointmentStore extends StartStopService implements IAppointmentSt
...
-            const currentAppointment = this.appointmentsByStateLocator[appointmentById.getStateLocator()];
-            // if it has the same id
-            if (currentAppointment.id === appointmentId) {
-                delete this.appointmentsByStateLocator[appointmentById.getStateLocator()];
-            }
-
-            // and remove from remote storage
-            await this.db.del(appointmentId);
-
+            await this.lockManager.withLock(appointmentById.getStateLocator(), async () => {
```
yahgwai: At the moment this isnt strictly necessary since the lock contents is blocking code. I'm tempted to say the lock should remain here anyway, as it might be easy for someone to put an await in here and forget to apply the lock. What are your thoughts?

src/watcher/store.ts
```diff
@@ -74,35 +77,37 @@ export class AppointmentStore extends StartStopService implements IAppointmentSt
...
-                        } nonce ${currentAppointment.getStateNonce()}`
-                    )
-                );
-
-                return false;
-            } else {
-                // remove the old appointment
-                delete this.appointmentsById[currentAppointment.id];
+    public addOrUpdateByStateLocator(appointment: IEthereumAppointment): Promise<boolean> {
+        return this.lockManager.withLock(appointment.getStateLocator(), async () => {
```
yahgwai: I think you should include a detailed comment here on why and when the state locator needs to be locked. Also on why the appointment id does not need to be locked.

test/src/utils/lock.test.ts
```diff
@@ -0,0 +1,179 @@
...
+        await lockManager.acquire("key");
+        let secondLockAcquired = false;
+        lockManager.acquire("key").then(() => {
+            secondLockAcquired = true;
+        });
+        await wait(30);
+        expect(secondLockAcquired).to.be.false;
+    });
+
+    it("acquire cannot acquire a key again before release", async () => {
```
yahgwai: This test has a duplicate name

src/utils/lock.ts
```diff
@@ -0,0 +1,100 @@
...
+     * Released the lock.
+     *
+     * @throws ApplicationError if the called when the lock was
+     */
+    public release() {
+        if (!this.mLocked) throw new ApplicationError("Tried to release a Lock that was not locked");
+
+        if (this.waiters.length > 0) {
+            // resolve the first waiter in the queue
+            const resolve = this.waiters.shift()!;
```
yahgwai: Why does this need the assertion !

src/watcher/store.ts
bigspider: But then, if there is an update to the state locator that leads to a db action adding the same appointmentId, we don't have a guarantee that the insertion happens after `db.del`; if not, it would be immediately deleted on the db, but not in memory.
Am I missing something?

src/utils/lock.ts
bigspider: Because shift() can return undefined when length === 0, and the type system can only infer non-null checks.

test/src/utils/lock.test.ts
bigspider: Removed, also the tested functionality was redundant (essentially equal to two tests before).

src/utils/lock.ts
```diff
@@ -0,0 +1,100 @@
...
+     *
+     * @param key
+     * @param func
+     */
+    public async withLock<T>(key: string, func: () => Promise<T>): Promise<T> {
+        try {
+            await this.acquire(key);
+            return await func();
+        } finally {
+            await this.release(key);
```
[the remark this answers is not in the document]
bigspider: Removed.

src/utils/lock.ts
```diff
@@ -0,0 +1,100 @@
...
+export class Lock {
+    private waiters: Array<() => void> = [];
+    private mLocked = false;
+
+    public get locked() {
+        return this.mLocked;
+    }
+
+    /**
+     * Acquires the lock; returns a promise that resolves when the lock is acquired.
```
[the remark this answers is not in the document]
bigspider: Done.

src/watcher/store.ts
bigspider: As per discussion, removed db.del from the critical section, and added a comment.

src/watcher/store.ts
bigspider: Left the lock, as discussed.

src/watcher/store.ts
bigspider: Done in 736c3f65ce697ca696ab734d873fbc468b4b0590.

PR #167, opened 2019-05-22
Added more robust block monitoring utils

bigspider:
Summary of changes:
- New classes: BlockCache, BlockProcessor, ConfirmationObserver, BlockTimeoutDetector in /blockMonitor.
- Refactored ReorgDetector to use the new BlockCache and BlockProcessor; changed reorg handling to be synchronous thanks to the BlockCache.
- Removed utilities in /utils/ethers.ts and their tests (replaced by the above services).
- Heavily refactored EthereumTransactionMiner and EthereumDedicatedResponder to use the new classes instead of the (with some marginal semantics changes)
- Some minor unrelated code cleanups included.


src/blockMonitor/blockCache.ts
```diff
@@ -0,0 +1,230 @@
...
+    /**
+     * Returns number of confirmations using `headBlockHash` as tip of the blockchain, looking for `txHash` among the ancestor blocks;
+     * return 0 if no ancestor containing the transaction is found.
+     * Note: This will return 0 for transactions already at depth bigger than `this.maxDepth` when this function is called.
+     */
+    public getConfirmations(headBlockHash: string, txHash: string): number {
+        let depth = 0;
+        let curBlock = this.getBlockStub(headBlockHash);
+        while (curBlock !== null) {
+            const txsInCurBlock = this.txHashesByBlockHash.get(curBlock.hash) || new Set();
```
yahgwai: very minor point, but instead of creating a new set you could check for null:
```
if (txsInCurBlock && txsInCurBlock.has(txHash)) {
                return depth + 1;
            }
```

src/blockMonitor/blockCache.ts
```diff
@@ -0,0 +1,230 @@
...
+ * It does not allow to add blocks without adding their parent first, except if they are at depth `maxDepth`.
+ **/
+export class BlockCache {
+    public blockStubsByHash: Map<string, BlockStubChain> = new Map();
+
+    // store block hashes at a specific height (there could be more than one at some height because of forks)
+    public blockHashesByHeight: Map<number, Set<string>> = new Map();
+
+    // maps transaction hashes to blocks containing them
+    public blockHashesByTxHash: Map<string, Set<string>> = new Map();
```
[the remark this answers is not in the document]
bigspider: Removed.

src/blockMonitor/blockCache.ts
```diff
@@ -0,0 +1,230 @@
...
+}
+
+/**
+ * Utility class to store and query info on full blocks up to a given maximum depth `maxDepth`, compared to the current
+ * maximum height ever seen.
+ * It prunes all the blocks at depth bigger than `maxDepth`, or with height smaller than the first block that was added.
+ * It does not allow to add blocks without adding their parent first, except if they are at depth `maxDepth`.
+ **/
+export class BlockCache {
+    public blockStubsByHash: Map<string, BlockStubChain> = new Map();
```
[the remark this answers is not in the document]
bigspider: Done.

src/blockMonitor/blockCache.ts
```diff
@@ -0,0 +1,230 @@
...
+                return false;
+            }
+
+            if (block.number < this.getMinVisibleHeight()) {
+                // block too deep
+                return false;
+            }
+
+            // If the block's parent is above the minimum visible height, it needs to be added first
+            if (!this.canAddBlock(block)) {
```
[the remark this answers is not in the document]
bigspider: I wanted to distinguish the situation where you try to add a block that is not necessary (i.e. would be pruned immediately, or block already added), from syntactically illegal attempts (parent should be added first). So I can return `false` for legal-but-redundant, and throw an exception for illegal attempts.
Lacking tests for this behavior, though.

src/blockMonitor/blockCache.ts
```diff
@@ -0,0 +1,230 @@
...
+
+        // Link all the tx hashes of this block to the block hash
+        for (let txHash of block.transactions || []) {
+            // Add this block among the block containing transaction txHash
+            addItemToKeyedSet(this.blockHashesByTxHash, txHash, block.hash);
+        }
+
+        // If the maximum block height increased, we might have to prune some old info
+        if (this.mMaxHeight < block.number) {
+            this.mMaxHeight = Math.max(this.mMaxHeight, block.number);
```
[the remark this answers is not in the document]
bigspider: You might add a new block to the cache without increasing the height, e.g. if you're filling the blocks on a different branch that is not the main chain.
Note that the BlockCache does _not_ keep track of the current valid chain, but it uses the height of the tallest block to decide when it can prune blocks. I think this is safe, as usually the currently valid chain should be the longest, although technically it's the maximum-work valid chain one and not the longest one.

src/blockMonitor/blockCache.ts
```diff
@@ -0,0 +1,230 @@
...
+
+    // Blocks at height smaller than which all blocks have already been pruned
+    private minStoredHeight = 0;
+
+    // After the first block is added, its height. Will not store blocks with smaller height
+    private initialHeight: number = -1;
+
+    // Height of the highest known block
+    private mMaxHeight = 0;
+    public get maxHeight() {
```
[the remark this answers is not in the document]
bigspider: Personally I would leave it there as harmless, as otherwise a user of the blockCache who was not the one adding blocks has no way of knowing about the current height of the cache.

src/blockMonitor/blockCache.ts
```diff
@@ -0,0 +1,230 @@
...
+        return this.mMaxHeight;
+    }
+
+    // Returns the minimum height of blocks that can currently be stored
+    public getMinVisibleHeight() {
```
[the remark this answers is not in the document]
bigspider: Similarly to the `maxHeight` field above, I think the class should have some getter to know the current boundary of the cache. Otherwise, if I have a certain block hash and height, I cannot distinguish getBlock(hash) is returning null because the block does not exists, or it was already pruned.
Perhaps this should become a `public get minVisibleHeight` for better naming consistency.

src/blockMonitor/blockCache.ts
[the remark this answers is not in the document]
yahgwai: Yeah, but at this point block.number is guaranteed to be greater than mMaxHeight as you've just checked that this.mMaxHeight < block.number. So Math.max(this.mMaxHeight, block.number) === block.number

src/blockMonitor/blockCache.ts
```diff
@@ -0,0 +1,230 @@
...
+
+        return true;
+    }
+
+    /**
+     * Returns the `BlockStubChain` for the block with hash `blockHash`, or `null` if the block is not in cache.
+     * @param blockHash
+     */
+    public getBlockStubChain(blockHash: string): BlockStubChain | null {
+        const blockStubChain = this.blockStubsByHash.get(blockHash);
```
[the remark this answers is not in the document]
bigspider: Done, short is very clear.

src/blockMonitor/blockCache.ts
```diff
@@ -0,0 +1,230 @@
...
+        return blockStubChain;
+    }
+
+    /**
+     * Returns the `IBlockStub` for the block with hash `blockHash`, or `null` if the block is not in cache.
+     * @param blockHash
+     */
+    public getBlockStub(blockHash: string): IBlockStub | null {
+        const blockStubChain = this.getBlockStubChain(blockHash);
+        if (blockStubChain === null) {
```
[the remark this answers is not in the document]
bigspider: Not a big fan of this very JavaScripty use of  &&, especially when used to call functions conditionally.

src/blockMonitor/blockCache.ts
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    /**
+     * Constructs a block forest that stores blocks up to a maximum depth `maxDepth`.
+     * @param maxDepth
+     */
+    constructor(public readonly maxDepth: number) {}
+
+    // Removes all info related to a block in blocksByHash and blockHashesByTxHash
+    private pruneBlock(blockHash: string) {
```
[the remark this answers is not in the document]
bigspider: Done.

src/blockMonitor/blockCache.ts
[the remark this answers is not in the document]
yahgwai: Yeah, fair enough

src/blockMonitor/blockCache.ts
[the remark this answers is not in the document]
yahgwai: OK, I see what you're getting at. You shouldn't put the canAddBlock in an 'else' branch though, although at the moment the logic is consistent it may not always be. canAddBlock should be check every time someone tries to add a block, without exception. Also, as a user I would probably call canAddBlock() before calling addBlock() everytime - at least that would seem the sensible thing to do, and at the moment canAddBlock contains the block already added + pruned immediately checks that you're talking about - so I would never get the false return value.

src/blockMonitor/blockCache.ts
```diff
@@ -0,0 +1,230 @@
+import { ethers } from "ethers";
+import { ApplicationError } from "../dataEntities";
+import { IBlockStub, BlockStubChain } from "./blockStub";
+
+// adds item to map.get(key), but make sure it exists first
+function addItemToKeyedSet<T, U>(map: Map<T, Set<U>>, key: T, item: U) {
```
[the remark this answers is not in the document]
bigspider: After simplifying the code a bit, the utility functions were needed much anymore, so I removed them in b771cb0c5543974a0f1750b40829a4f52580011f.

src/blockMonitor/blockCache.ts
bigspider: Done.

src/blockMonitor/blockCache.ts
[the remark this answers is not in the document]
bigspider: Right, removed it.

src/blockMonitor/blockCache.ts
```diff
@@ -0,0 +1,230 @@
...
+    public blockHashesByTxHash: Map<string, Set<string>> = new Map();
+
+    // set of tx hashes per block hash, for fast lookup
+    public txHashesByBlockHash: Map<string, Set<string>> = new Map();
+
+    // Blocks at height smaller than which all blocks have already been pruned
+    private minStoredHeight = 0;
+
+    // After the first block is added, its height. Will not store blocks with smaller height
+    private initialHeight: number = -1;
```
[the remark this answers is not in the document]
bigspider: The initial height is also stored because it overrieds the "valid range" of the cache, that is, the valid range is always bounded by h := Max(initialHeight, maxHeight - depth) (that is, any new block needs to have heigth >= h).
I was using the value "-1" to mark if it is still empty; I agree on having a separate boolean variable for that purpose. After that, I renamed the former "minStoredHeight" to "pruneHeight" and it turns out that it was now redundants to have both `pruneHeight` and `initialHeight`, so I left only one.
See the changes in 44a72c72794afd9765aeadb305a8bbc959293140.

src/blockMonitor/blockCache.ts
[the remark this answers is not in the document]
bigspider: Renamed to the simpler "minHeight" in 44a72c72794afd9765aeadb305a8bbc959293140, consistent with maxHeight above.
I'll add some tests to verify correct behavior.

src/blockMonitor/blockCache.ts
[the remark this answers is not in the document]
bigspider: For now, moved the check at the beginning of the function in 44a72c72794afd9765aeadb305a8bbc959293140; much cleaner looking, indeed.

Still thinking if the semantic can be simplified a bit.

src/blockMonitor/blockCache.ts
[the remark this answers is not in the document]
bigspider: Added more tests in 401febdc98e9b4181999335826093859a13ea211.

src/blockMonitor/blockTimeoutDetector.ts
```diff
@@ -0,0 +1,58 @@
...
+        this.initNoNewBlockTimeout();
+    }
+
+    public async stopInternal(): Promise<void> {
+        this.clearNoNewBlockTimeout();
+        this.blockProcessor.off(BlockProcessor.NEW_HEAD_EVENT, this.handleNewBlock);
+    }
+
+    // Start the timer, after clearing any previous timer
+    private initNoNewBlockTimeout() {
```
yahgwai: I think setNoNewBlockTimeout is a slightly clearer name

src/blockMonitor/blockTimeoutDetector.ts
```diff
@@ -0,0 +1,58 @@
...
+            clearTimeout(this.timeoutHandler);
+            this.timeoutHandler = null;
+        }
+    }
+
+    private handleNoNewBlockTimeout() {
+        this.emit(BlockTimeoutDetector.BLOCK_TIMEOUT_EVENT);
+    }
+
+    private handleNewBlock() {
```
yahgwai: you dont need this function

src/blockMonitor/blockTimeoutDetector.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    // If a timer was active, clear it
+    private clearNoNewBlockTimeout() {
+        if (this.timeoutHandler !== null) {
+            clearTimeout(this.timeoutHandler);
+            this.timeoutHandler = null;
+        }
+    }
+
+    private handleNoNewBlockTimeout() {
```
yahgwai: Maybe just inline this with an arrow function and avoid the bind
```
setTimeout(() => this.emit(BlockTimeoutDetector.BLOCK_TIMEOUT_EVENT), this.timeout);
```

src/blockMonitor/reorgDetector.ts
```diff
@@ -110,6 +116,11 @@ export class ReorgDetector extends StartStopService {
      * @param newHead
      */
     private async conductReorg(newHead: BlockStubChain) {
+        // We ignore further reorgs until this one is complete
+        if (this.conductingReorg) return;
```
yahgwai: I dont think this is safe - if two reorgs happen concurrently they will both update the head block but only one of them will fire relevant reorg events

src/blockMonitor/reorgDetector.ts
```diff
@@ -131,18 +142,17 @@ export class ReorgDetector extends StartStopService {
...
     }
 
     /**
      * Since this reorg detector only detects reorgs below a max depth it can prune records
      * that it has below that
      */
     private prune() {
-        // prune the chain
-        const minHeight = this.headBlock.height - this.maxDepth;
-        this.headBlock.prune(minHeight);
```
yahgwai: What happened to this line?

src/blockMonitor/reorgDetector.ts
```diff
@@ -222,20 +232,15 @@ export class ReorgDetector extends StartStopService {
...
+            commonAncestor = this.findCommonAncestorDeep(newBlock.parentHash, currentHead, differenceBlocks, minHeight);
         }
 
         return { commonAncestor, differenceBlocks };
     }
 
     /**
      * Add a listener for reorg events that reorg the chain to a common ancestor below a certain height. These events are guaranteed
-     * to fire after ReorgDetector.REORG_START_EVENT and before ReorgDetector.REORG_END_EVENT
+     * to fire after ReorgDdetector.REORG_START_EVENT and before ReorgDetector.REORG_END_EVENT
```
yahgwai: Typo ReorgDdetector

src/blockMonitor/blockTimeoutDetector.ts
yahgwai: This should be wrapped in a try/catch

src/responder.ts
```diff
@@ -201,77 +204,67 @@ export class EthereumTransactionMiner {
...
 
         try {
             // First, wait to get at least 1 confirmation, but throw an error if the transaction is stuck
             // (that is, new blocks are coming, but the transaction is not included)
-            await Promise.race([
-                firstConfirmationPromise,
-                firstConfirmationTimeoutPromise,
-                noNewBlockPromise
-            ]);
+            await Promise.race([firstConfirmationPromise, noNewBlockPromise]);
```
yahgwai: Doesnt the noNewBlock promise need to be cancelled?

src/responder.ts
```diff
@@ -201,77 +204,67 @@ export class EthereumTransactionMiner {
...
         try {
             // Then, wait to get at enough confirmations; now only throw an error if there is a reorg
-            await Promise.race([
-                enoughConfirmationsPromise,
-                noNewBlockPromise
-            ]);
+            await Promise.race([enoughConfirmationsPromise, noNewBlockPromise]);
         } finally {
             // Make sure any pending CancellablePromise is released
             enoughConfirmationsPromise.cancel();
```
yahgwai: Again, doesnt the noNewBlockPromise need to be cancelled?

src/responder.ts
```diff
@@ -417,57 +382,92 @@ export class EthereumDedicatedResponder extends EthereumResponder {
...
+    constructor(private readonly signer: ethers.Signer, blockCache: BlockCache, blockProcessor: BlockProcessor) {
+        super("EthereumResponderManager");
+
         if (!signer.provider) throw new ArgumentError("The given signer is not connected to a provider");
 
         this.provider = signer.provider;
 
         this.gasPolicy = new DoublingGasPolicy(this.provider);
+
+        this.blockTimeoutDetector = new BlockTimeoutDetector(blockProcessor, 120 * 1000);
```
yahgwai: These should be detected, and EthereumResponderManager shouldnt be a startstopservice

src/watcher/store.ts
```diff
@@ -42,7 +42,7 @@ export class AppointmentStore extends StartStopService implements IAppointmentSt
         super("Appointment store");
     }
 
-    public async startInternal() {
+    protected async startInternal() {
```
yahgwai: Nice catch :+1:

test/src/responder.test.ts
```diff
@@ -303,119 +342,100 @@ describe("EthereumDedicatedResponder", () => {
...
+        expect(args[1] instanceof BlockTimeoutError, "AttemptFailed emitted with BlockTimeoutError").to.be.true;
     });
 
     // TODO: this test currently passes but node complains with UnhandledPromiseRejectionWarning for Reorg (but exception is caught in the responder!)
     it("emits the AttemptFailed with a ReorgError if a re-org kicks out the transaction before enough confirmations", async () => {
-        const { signer, appointment, responseData } = this.testData;
+        const { signer, appointment, responseData } = testData;
+
+        const mockedTransactionMiner = mock(EthereumTransactionMiner);
+        when(mockedTransactionMiner.sendTransaction(anything())).thenResolve("0x1234");
```
yahgwai: :+1:

test/src/watcher/watcher.test.ts
```diff
@@ -83,6 +84,8 @@ describe("Watcher", () => {
     when(mockedReorgDetector.addReorgHeightListener(anyNumber(), anything())).thenReturn();
     const reorgDetectorInstance = instance(mockedReorgDetector);
 
+    const mockedBlockProcessor = mock(BlockProcessor);
```
yahgwai: Seems like a left over from refactoring, not needed any more

src/blockMonitor/confirmationObserver.ts
```diff
@@ -0,0 +1,130 @@
+import { StartStopService, ArgumentError, BlockThresholdReachedError, ReorgError } from "../dataEntities";
```
yahgwai: As I mentioned before, I think the logic in this class is a bit tricky to understand. Try refactoring it to make it clearer, the main thing I'm thinking about is that the class does too much in a single function. Start by separating the functionality, then you can find a nice way of recombining them afterwards to make it easy for a caller - or maybe leave the separate for more flexibility.

src/blockMonitor/blockProcessor.ts
```diff
@@ -0,0 +1,69 @@
...
+    private async handleBlockEvent(blockNumber: number) {
+        const observedBlock = await this.provider.getBlock(blockNumber);
+
+        this.lastBlockHashReceived = observedBlock.hash;
+
+        // populate block and parents in cache
+        if (!this.blockCache.hasBlock(observedBlock.hash)) {
+            (await this.getParentsNotInCache(observedBlock))
+                .concat(observedBlock)
+                .filter(b => !this.blockCache.hasBlock(b.hash))
```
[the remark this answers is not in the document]
bigspider: I think it was semantically equivalent in this case, but I rewrote the whole function in a much simpler way in d0d4db0f71a37e74668fbcb9206ed67710412904.

src/blockMonitor/blockProcessor.ts
```diff
@@ -0,0 +1,69 @@
...
+            return [];
+        } else {
+            // this parent is not in the cache, find further parents also in this situation
+            const parentBlock = await this.provider.getBlock(block.parentHash);
+            return (await this.getParentsNotInCache(parentBlock)).concat(parentBlock);
+        }
+    }
+
+    private async handleBlockEvent(blockNumber: number) {
+        const observedBlock = await this.provider.getBlock(blockNumber);
```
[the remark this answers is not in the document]
bigspider: Actually, without `true` block.transactions is an array of transaction hashes; with `true` it is an array of [transaction responses](https://docs.ethers.io/ethers.js/html/api-providers.html#transaction-response). This is quite annoying as the documentation does not seem to specify this anywhere :/

src/blockMonitor/blockProcessor.ts
```diff
@@ -0,0 +1,69 @@
...
+            // this parent is in the cache - do nothing
+            return [];
+        } else {
+            // this parent is not in the cache, find further parents also in this situation
+            const parentBlock = await this.provider.getBlock(block.parentHash);
+            return (await this.getParentsNotInCache(parentBlock)).concat(parentBlock);
+        }
+    }
+
+    private async handleBlockEvent(blockNumber: number) {
```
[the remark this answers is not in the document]
bigspider: Done in c296e46148580b4d640213b0d219e648a00e9a64.

src/blockMonitor/blockStub.ts
```diff
@@ -47,7 +52,7 @@ export class BlockStubChain {
                 throw new ArgumentError("Parent hashes are not equal.", extensionBlock.parentHash, block.hash);
```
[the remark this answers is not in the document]
bigspider: Done in c2a40526b5ed6d5450775d04346664d8f1b64942.

src/blockMonitor/blockTimeoutDetector.ts
bigspider: Why? It cannot throw an exception, as far as I can tell. It is only clearing a timeout handler (if any), and calling a new setTimeout

Renamed to **reset**NoNewBlockTimeout, since it possibly also clears the existing timeout. (f862b17adf715ae2db894a21a4292a1ac9912b62)

src/blockMonitor/blockTimeoutDetector.ts
bigspider: Removed in f862b17adf715ae2db894a21a4292a1ac9912b62.

src/blockMonitor/blockTimeoutDetector.ts
bigspider: The whole line becomes
```
this.timeoutHandler = setTimeout(() => this.emit(BlockTimeoutDetector.BLOCK_TIMEOUT_EVENT), this.timeout);
```
which I find quite hard to read.

yahgwai: Ahh, I didnt think it was too complex. I thought you wouldnt either, as this is very similar to the Promise syntax for injecting an inline function. For example we inject some large functions into CancellablePromise:

If you prefer, maybe set the function to a variable on the line above then pass it in.
```
const blockEventHandler = () => this.emit(BlockTimeoutDetector.BLOCK_TIMEOUT_EVENT);
this.timeoutHandler = setTimeout(blockEventHandler, this.timeout);

```

src/blockMonitor/blockTimeoutDetector.ts
yahgwai: Yeah, I think we had a chat about this once before. I feel like it's error prone either way.

If you don't add a try/catch someone may to edit your logic, make it more complex and not think to add the try/catch which they now need to do.

If you add a try/catch it's unnecessary so someone it's likely to just remove it, even if you comment it. In which case you're back at square one. In the mean time, you've also added unnecessary code.

What I was kinda thinking was that we could just make a manual global rule which is: if it's an event handler, wrap it in try/catch. But it's a human process, so it probably wont work either. What dyou think? I also don't mind if you don't add the try/catch, since none of the above solutions are good ones.

bigspider: Yeah, generally speaking, I would say that _all_ code should catch potential exceptions, if they can happen; it would be good if that could be enforced at compile-level, but we're not that lucky with TypeScript.
One option could be to use a decorator for all event handlers to handle the logic of wrapping in a try/catch and logging any error, but they are still an experimental feature, not sure we want to use them: https://www.typescriptlang.org/docs/handbook/decorators.html

src/blockMonitor/blockTimeoutDetector.ts
bigspider: Ok, done it in af09fd3195046d572ab6ca75b3a3abaa5acd058c.

src/blockMonitor/reorgDetector.ts
bigspider: You're right; changed it in e31cb01 to stop processing new blocks instead. It still feels unsatisfactory, but I guess further improvemens could be part of #168.

src/blockMonitor/reorgDetector.ts
bigspider: I'm not sure why this is here, but those lines are still there in the current version of the code (in reorgDetector.ts instead of reorg.ts).

src/blockMonitor/blockTimeoutDetector.ts
yahgwai: I like the idea of a decorator as it would be declare it's use case: @EventHandler

We can do this/investigate in a different ticket for now. Do what you want for the case at hand.

src/blockMonitor/reorgDetector.ts
yahgwai: The following line appears to be missing in reorgDetector.ts.
```
this.headBlock.prune(minHeight);
```

src/blockMonitor/reorgDetector.ts
yahgwai: Yeah, I agree that this still looks a bit weird. But from what I can see it's safe now. I think we should change the reorg detector logic quite drastically soon, so as long as it's safe lets leave it for now.

src/blockMonitor/reorgDetector.ts
bigspider: It's only pruning the store for the height events:
```
    /**
     * Since this reorg detector only detects reorgs below a max depth it can prune records
     * that it has below that
     */
    private prune() {
        // prune current re-org height listeners
        const minHeight = this.headBlock.height - this.maxDepth;
        this.store.prune(minHeight);
    }
```

src/blockMonitor/reorgDetector.ts
bigspider: Fixed.

src/responder.ts
bigspider: Solved in ffc3e167ad041682416a75ae73c07cb439831e5b

src/responder.ts
bigspider: Solved in ffc3e167ad041682416a75ae73c07cb439831e5b.

test/src/watcher/watcher.test.ts
bigspider: Removed.

src/responder.ts
bigspider: Done in fcfe337.

src/blockMonitor/confirmationObserver.ts
bigspider: Simplified quite a lot in 11331a6e57a354cc70f69f81cbaff4c5fcf0237f, by splitting waitForConfirmations into two separate functions: waitForFirstConfirmation (that optionally throws if the transaction is stuck, but does not care about reorgs), and waitForConfirmations(that optionally throws for Reorg, but assumes that the transaction was already confirmed).

There is still a bit too much CancellablePromise-related boilerplate, but I think it is manageable, and much easier to read now.

PR #170, opened 2019-05-27
Added existing appointments to subscriber upon startup of watcher

yahgwai:


test/src/watcher/watcher.test.ts
```diff
@@ -284,6 +285,33 @@ describe("Watcher", () => {
...
+        ).once();
+        verify(
+            mockedAppointmentSubscriber.subscribe(
+                appointmentNotUpdated.id,
+                appointmentCanBeUpdated.getEventFilter(),
+                anything()
+            )
+        ).once();
+        await watcher.start();
     });
```
bigspider: This should be `.stop()`.

yahgwai: woops, updated.

PR #177, opened 2019-05-29
Added per-service named loggers

bigspider:


src/dataEntities/startStop.ts
```diff
@@ -18,13 +19,20 @@ export abstract class StartStopService extends EventEmitter {
...
+
     /**
      * A service that requires starting and stopping.
      * Whoever constructs this service must start it before using it.
      * Whoever constructs this service must stop it after using it.
      */
     protected constructor(protected readonly name: string) {
         super();
+
+        this.logger = createNamedLogger(name);
```
yahgwai: The names currently being used are of inconsistent formats. I think we should standardise them - documentation is enough. Something along the lines of "Names must be lower case, words separated by hypthons".

We can do a quick run time check that this is the case, and throw a configuration if necessary.

The problem with using the names as they are now is that they have spaces in them - which isnt nice for file names.

src/dataEntities/startStop.ts
```diff
@@ -41,7 +49,7 @@ export abstract class StartStopService extends EventEmitter {
         // set started straight away to block the code below
         this.mStarting = true;
         await this.startInternal();
-        logger.info(`${this.name}: Started.`);
+        this.logger.info(`${this.name}: Started.`);
```
yahgwai: You no longer need ${this.name} in any logs of this class, or of it's sub classes

src/watcher/garbageCollector.ts
```diff
@@ -84,9 +83,9 @@ export class AppointmentStoreGarbageCollector extends StartStopService {
                 // an error here means that we were likely unable to collect all, or some of, the appointments
                 // consecutive errors could mean we have a systematic problem, or connection issues with the store
                 // in either case the problem is very serious - so we stop the GC
-                logger.error("GC: Unexpected error.");
-                logger.error(`GC: Consecutive errors: ${this.consecutiveErrors}.`);
-                logger.error(doh);
+                this.logger.error("GC: Unexpected error.");
```
yahgwai: Remove the "GC" string from the logs in this class

src/logger.ts
```diff
@@ -15,22 +15,31 @@ const myFormat = format.printf(info => {
...
+
+export function createNamedLogger(name: string | null) {
+    const prefix = name !== null ? name + "-" : "";
+    const newLogger = createLogger({
+        level: "info",
+        format: combinedFormats,
+        transports: [
+            new transports.File({ dirname: logDir, filename: `${prefix}error.log`, level: "error" }),
+            new transports.File({ dirname: logDir, filename: `${prefix}info.log`, level: "info" }),
+            new transports.File({ dirname: logDir, filename: `${prefix}debug.log`, level: "debug" })
```
yahgwai: I'm not sure why, but this doesnt appear to be working. Logs are appearing for debug - but they only contain info messages. Not sure if this was introduced by this PR, might have been a bug already. If so raise a bug ticket for it separately.

package.json
```diff
@@ -12,12 +12,12 @@
     "start-evm": "ganache-cli -d -l 15000000 --allowUnlimitedContractSize --defaultBalanceEther 1000000 --mnemonic 'myth like bonus scare over problem client lizard pioneer submit female collect'",
     "submodule": "git submodule init && git submodule update",
     "lint": "eslint --ext ts .",
-    "test": "NODE_ENV=test && mocha -r ts-node/register 'test/src/**/*.test.ts' --exit && npm run test-mocked",
+    "test": "export NODE_ENV=test; mocha -r ts-node/register 'test/src/**/*.test.ts' --exit && npm run test-mocked",
```
yahgwai: Did you find these exports were necessary in the end?

bigspider: I investigated further, it seems that "NODE_ENV=test <command>" works, but not "NODE_ENV=test && command"; with export both versions work, which is why I was getting more consistent behavior, as we had a mix.
I removed the export and the also the extra &&.

src/dataEntities/startStop.ts
bigspider: Done, using the regexp `/^[a-z0-9\-]+$/`, so letters, numbers and hyphens.

src/dataEntities/startStop.ts
bigspider: Removed.

src/watcher/garbageCollector.ts
bigspider: Removed.

src/logger.ts
bigspider: It seems to be working as intended to me: "-error" only contains errors, "-info" contains errors and infos, while "-debug" contains everything. (see for example `watcher-debug.log` that has all the three levels).

yahgwai: Yeah, that's true, I was conflating two things.

The other thing then is that we appear to logging with level: "info" - so why should the debug files be created at all?

src/dataEntities/startStop.ts
yahgwai: It's good feedback to include the offending name as part of the error message.

src/logger.ts
bigspider: Not sure I get what you mean. "debug" is the most verbose level, and we currently don't have it in many places, but the watcher is using it to show detailed Event information, which is about 12 lines of output.
Or would you rather only have "error" and "info" to reduce the number of log files?

src/dataEntities/startStop.ts
bigspider: Added better feedback in b083c9975182a41a572013f1993a80357e4fb63f.

src/logger.ts
yahgwai: Ah no, it's fine to have debug. I'm just surprised that the logs are showing up.

```
createLogger({
level: "info",
format: combinedFormats,
```

I've looked up the documentation for `logger: info` and apparently that's supposed to be the minimum that's being logged. So I was wondering why debug files were being created when we have the minimum set to 'info' above.

yahgwai: Anyways, I'm sure this is a separate issue from your PR

PR #181, opened 2019-06-03
Add option for log verbosity, and default to verbose logs for tests

bigspider:


src/logger.ts
```diff
@@ -2,6 +2,27 @@ import { createLogger, format, transports } from "winston";
 import { getRequestId } from "./customExpressHttpContext";
 import fs from "fs";
 const logDir = "logs";
+
+// For now, we only support three log levels
+export type LogLevel = "error" | "info" | "debug";
+export const supportedLogLevels: LogLevel[] = ["error", "info", "debug"];
```
yahgwai: You can restructure these repeated strings by having a LogLevelInfo class which could look something like:
```
enum LogLevel {
Info = "info",
Error = "error",
Debug = "debug"
}

class LogLevelInfo {

public static Error : LogLevelInfo = new LogLevelInfo(LogLevel.Error, 0);
public static Info : LogLevelInfo = new LogLevelInfo(LogLevel.Info, 2);
public static Debug : LogLevelInfo = new LogLevelInfo(LogLevel.Debug, 4);

private constructor(public readonly logLevel: LogLevel, public readonly order: number) {}

static tryParse(logLevel: string):LogLevelInfo { 
// returns null if level does not exist
}
}

Comment (bigspider) [reply src/logger.ts]: Refactored with Enums; clean enough, but it doesn't seem to be possible to stay fully type-safe, unfortunately :/

File: src/startUp.ts [hunk 2]
```diff
@@ -26,12 +26,19 @@ const checkArgs = (args: IArgConfig) => {
         console.error("Options 'rate-limit-global-windowms' and 'rate-limit-global-max' must be provided together.");
         process.exit(1);
     }
+
+    if (commandLineConfig.loglevel && !Object.values(LogLevel).includes(commandLineConfig.loglevel)) {
```
yahgwai: This should use tryParse

src/startUp.ts
```diff
@@ -26,12 +26,19 @@ const checkArgs = (args: IArgConfig) => {
...
+        console.error("Options 'loglevel' can only be 'error', 'info' or 'debug'.");
+        process.exit(1);
+    }
 };
 
 async function startUp() {
     checkArgs(commandLineConfig);
     const config = Object.assign(jsonConfig, commandLineConfig);
 
+    setLogLevel(config.loglevel as LogLevel);
```
yahgwai: After using tryParse above, you will no longer need this cast.

src/startUp.ts
bigspider: it gets a bit awkward, though, as tryParse returns a LogLevelInfo instance. Then I should generalize setLogLevel to work with both LogLevel or LogLevelInfo.

yahgwai: just use the property on LogLevelInfo

logLevelInfo.Level

src/startUp.ts
bigspider: Well, config is IArgConfig, so config.loglevel is still just a string.
Instead, I changed setLogLevel to accept a string and throw ApplicationError if a wrong level is passed.

src/startUp.ts
bigspider: Changed with TryParse, and also changed the signature of setLogLevel (see other comment below).

PR #182, opened 2019-06-04
Make the BlockCache accessible via the interface of BlockProcessor; make sure that the head of the blockProcessor is always initialized

bigspider:
Changes in this PR:
- Split the BlockCache from its read-only interface (ReadOnlyBlockCache)
- Made the read-only view of the BlockCache accessible as a member of BlockProcessor
- The blockProcessor now downloads the last block from the provider, so that blockProcessor.head is never null after successfully running start()
- Incidentally, generalized waitForEvent from /utils in order to work also with classes that are not strictly an EventEmitter like the BaseProvider; moreover, it now resolves to the array of arguments passed to the event.

### DIFF
--- a/src/blockMonitor/blockCache.ts
+++ b/src/blockMonitor/blockCache.ts
@@ -3,6 +3,20 @@ import { ApplicationError } from "../dataEntities";
 import { IBlockStub, BlockStubChain } from "./blockStub";
 
 /**
+ * This interface represents the read-only view of a BlockCache.
+ */
+export interface ReadOnlyBlockCache {
+    readonly maxDepth: number;
+    readonly maxHeight: number;
+    readonly minHeight: number;
+    canAddBlock(block: ethers.providers.Block): boolean;
+    getBlockStubChain(blockHash: string): BlockStubChain | null;
+    getBlockStub(blockHash: string): IBlockStub | null;
+    hasBlock(blockHash: string): boolean;
+    getConfirmations(headBlockHash: string, txHash: string): number;
+}
+
+/**
  * Utility class to store and query info on full blocks up to a given maximum depth `maxDepth`, compared to the current
  * maximum height ever seen.
  * It prunes all the blocks at depth bigger than `maxDepth`, or with height smaller than the first block that was added.
@@ -19,23 +33,23 @@ import { IBlockStub, BlockStubChain } from "./blockStub";
  * Note that in order to guarantee the invariant (1), `addBlock` can be safely called even for blocks that will not
  * actually be added (for example because they are already too deep); in that case, it will return `false`.
  **/
-export abstract class ReadOnlyBlockCache {
-    protected blockStubsByHash: Map<string, BlockStubChain> = new Map();
+export class BlockCache implements ReadOnlyBlockCache {
+    private blockStubsByHash: Map<string, BlockStubChain> = new Map();
 
     // set of tx hashes per block hash, for fast lookup
-    protected txHashesByBlockHash: Map<string, Set<string>> = new Map();
+    private txHashesByBlockHash: Map<string, Set<string>> = new Map();
 
     // store block hashes at a specific height (there could be more than one at some height because of forks)
-    protected blockHashesByHeight: Map<number, Set<string>> = new Map();
+    private blockHashesByHeight: Map<number, Set<string>> = new Map();
 
     // Next height to be pruned; the cache will not store a block with height strictly smaller than pruneHeight
-    protected pruneHeight: number;
+    private pruneHeight: number;
 
     // True before the first block ever is added
-    protected isEmpty = true;
+    private isEmpty = true;
 
     // Height of the highest known block
-    protected mMaxHeight = 0;
+    private mMaxHeight = 0;
     public get maxHeight() {
         return this.mMaxHeight;
     }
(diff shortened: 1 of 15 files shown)


src/blockMonitor/blockProcessor.ts
```diff
@@ -10,10 +11,30 @@ import { IBlockStub } from "./blockStub";
...
+
+    /**
+     * Returns the ReadOnlyBlockCache associated to this BlockProcessor.
+     */
+    public get blockCache(): ReadOnlyBlockCache {
+        return this.mBlockCache;
+    }
+
+    /**
+     * Returns the IBlockStub of the latest known head block, or null if none is known yet
```
yahgwai: Comment is out of date

src/blockMonitor/blockProcessor.ts
```diff
@@ -23,21 +44,27 @@ export class BlockProcessor extends StartStopService {
...
+        this.mBlockCache = blockCache;
+
+        this.processBlockNumber = this.processBlockNumber.bind(this);
     }
 
     protected async startInternal(): Promise<void> {
-        this.provider.on("block", this.handleBlockEvent);
+        // Make sure the current head block is processed
+        const initialBlockNumber = await this.provider.getBlockNumber();
+        await this.processBlockNumber(initialBlockNumber);
```
yahgwai: This currently emits an event on startup. Should we do this? Can we be sure that subscribers have all subscribed by the time this is called? And presumably other services will be starting up at around the same time, so the system may not full be intitialised. Maybe we should set `initialBlockNumber = await this.provider.getBlockNumber() - 1`, and not emit an event

src/blockMonitor/confirmationObserver.ts
```diff
@@ -59,20 +58,21 @@ export class ConfirmationObserver extends StartStopService {
...
+        const blockCache = this.blockProcessor.blockCache;
+        if (nConfirmations > blockCache.maxDepth) {
             return new CancellablePromise((_, reject) =>
                 reject(new ArgumentError("nConfirmations cannot be bigger than the BlockCache's maxDepth."))
             );
         }
 
         const headHash = this.blockProcessor.head && this.blockProcessor.head.hash;
-        if (headHash && this.blockCache.getConfirmations(headHash, txHash) >= nConfirmations) {
+        if (headHash && blockCache.getConfirmations(headHash, txHash) >= nConfirmations) {
```
yahgwai: Is this check on head hash necessary? And the one on the line above?

src/blockMonitor/blockProcessor.ts
```diff
@@ -1,7 +1,8 @@
 import { ethers } from "ethers";
-import { StartStopService } from "../dataEntities";
-import { BlockCache } from "./blockCache";
+import { StartStopService, ApplicationError } from "../dataEntities";
+import { ReadOnlyBlockCache, BlockCache } from "./blockCache";
 import { IBlockStub } from "./blockStub";
+import { waitForEvent } from "../utils";
```
yahgwai: Is this used?

src/blockMonitor/blockProcessor.ts
```diff
@@ -10,10 +11,30 @@ import { IBlockStub } from "./blockStub";
...
+    public get blockCache(): ReadOnlyBlockCache {
+        return this.mBlockCache;
+    }
+
+    /**
+     * Returns the IBlockStub of the latest known head block, or null if none is known yet
+     */
+    public get head(): IBlockStub {
+        const blockStub = this.blockCache.getBlockStub(this.headHash);
+        if (blockStub === null) {
```
yahgwai: Could be a truthy check

src/blockMonitor/blockProcessor.ts
yahgwai: We should probably be actually storing the block numbers so we know where to pick up from - but I guess that's another ticket

src/blockMonitor/confirmationObserver.ts
yahgwai: There is also a check in waitForBlocks that can be removed

src/blockMonitor/blockProcessor.ts
bigspider: I ended up not using it, instead. Removed the import.

src/blockMonitor/blockProcessor.ts
bigspider: Updated.

src/blockMonitor/confirmationObserver.ts
bigspider: Guard removed, and it revealed a couple of tests that needed fixing due to mocking the BlockProcessor incorrectly (fixed in 82087c8d41e383cd2c3287cd87efc96da92dbb1a).

src/blockMonitor/blockProcessor.ts
bigspider: Yep, fixed.

src/blockMonitor/blockCache.ts
```diff
@@ -58,7 +58,7 @@ export class BlockCache {
     constructor(public readonly maxDepth: number) {}
 
     // Removes all info related to a block in blockStubsByHash and txHashesByBlockHash
-    private removeBlock(blockHash: string) {
+    protected removeBlock(blockHash: string) {
```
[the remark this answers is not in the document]
bigspider: Yep, good observations. I thought doing it with an interface would be more painful than it actually was - done in 06fed892b8985de3ffc8e2d680cfe2c67ff5dde4.

src/blockMonitor/blockProcessor.ts
bigspider: Good catch, I thought that the event would not be emitted. I added an additional guard to only emit the event if the service is already started.

PR #192, opened 2019-06-20
Refactoring of the BlockCache and BlockProcessor

bigspider:
Refactored BlockProcessor and BlockCache in order to:
- allow type-safe extensibility of the base IBlockStub type
- further separate the ethers.js types from the block cache (now only the block processor is aware of ethers.js)

Also removed the getConfirmations functionality from the BlockCache (and the related internal dictionary) and re-implemented it as an external functionality.

Fixed a test in EthereumTransactionMiner that was broken due to imperfect mocking.

### DIFF
--- a/src/blockMonitor/blockProcessor.ts
+++ b/src/blockMonitor/blockProcessor.ts
@@ -2,13 +2,13 @@ import { ethers } from "ethers";
 import { StartStopService, ApplicationError } from "../dataEntities";
 import { ReadOnlyBlockCache, BlockCache } from "./blockCache";
 import { IBlockStub } from "../dataEntities";
-import { Block, HasTxHashes } from "../dataEntities/block";
+import { Block, Transactions } from "../dataEntities/block";
 
 type BlockFactory<T> = (provider: ethers.providers.Provider) => (blockNumberOrHash: number | string) => Promise<T>;
 
-export const minimalBlockFactory = (provider: ethers.providers.Provider) => async (
+export const blockStubAndTxFactory = (provider: ethers.providers.Provider) => async (
     blockNumberOrHash: string | number
-): Promise<IBlockStub & HasTxHashes> => {
+): Promise<IBlockStub & Transactions> => {
     const block = await provider.getBlock(blockNumberOrHash);
     return {
         hash: block.hash,
@@ -18,13 +18,15 @@ export const minimalBlockFactory = (provider: ethers.providers.Provider) => asyn
     };
 };
 
-export const defaultBlockFactory = (provider: ethers.providers.Provider) => async (
+export const blockFactory = (provider: ethers.providers.Provider) => async (
     blockNumberOrHash: string | number
 ): Promise<Block> => {
     const block = await provider.getBlock(blockNumberOrHash);
+
+    // We could filter out the logs that we are not interesting in order to save space
+    // (e.g.: only keep the logs from the DataRegistry).
     const logs = await provider.getLogs({
         blockHash: block.hash
-        // TODO: filter interesting logs, instead of taking everything
     });
 
     return {
--- a/src/blockMonitor/blockProcessor.ts
+++ b/src/blockMonitor/blockProcessor.ts
@@ -1,7 +1,40 @@
 import { ethers } from "ethers";
 import { StartStopService, ApplicationError } from "../dataEntities";
 import { ReadOnlyBlockCache, BlockCache } from "./blockCache";
-import { IBlockStub } from "./blockStub";
+import { IBlockStub } from "../dataEntities";
+import { Block, HasTxHashes } from "../dataEntities/block";
+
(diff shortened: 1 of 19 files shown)


src/dataEntities/block.ts
```diff
@@ -0,0 +1,17 @@
+import { ethers } from "ethers";
+
+export interface IBlockStub {
+    hash: string;
+    number: number;
+    parentHash: string;
+}
+
+export interface HasLogs {
```
yahgwai: I would just call LogList, as that's what it is. Same with HasTxHashes.

There's a TODO here that should be cleaned up. I would say stick with the ethers class for now.

src/blockMonitor/blockProcessor.ts
```diff
@@ -1,27 +1,62 @@
...
+    };
+};
+
+export const defaultBlockFactory = (provider: ethers.providers.Provider) => async (
+    blockNumberOrHash: string | number
+): Promise<Block> => {
+    const block = await provider.getBlock(blockNumberOrHash);
+    const logs = await provider.getLogs({
+        blockHash: block.hash
+        // TODO: filter interesting logs, instead of taking everything
```
yahgwai: TODO: still here

src/blockMonitor/blockTimeoutDetector.ts
```diff
@@ -16,7 +16,7 @@ export class BlockTimeoutDetector extends StartStopService {
     /**
      * @param timeout The number of milliseconds without a new block before generating
      */
-    constructor(private blockProcessor: BlockProcessor, public readonly timeout: number) {
+    constructor(private blockProcessor: BlockProcessor<any>, public readonly timeout: number) {
```
yahgwai: I think you should choose a type here instead of any

src/blockMonitor/confirmationObserver.ts
```diff
@@ -1,17 +1,19 @@
...
 
 /**
  * Allows to observe transactions to be notified when they reach a given number of confirmations.
  */
-export class ConfirmationObserver extends StartStopService {
+export class ConfirmationObserver<T extends IBlockStub & HasTxHashes> extends StartStopService {
     private txListeners = new Set<TransactionListener>();
 
-    constructor(private readonly blockProcessor: BlockProcessor) {
+    constructor(private readonly blockProcessor: BlockProcessor<T>) {
```
yahgwai: I dont think you should choose T here - just IBlockStub & HasTxHashes, that's all that ConfirmationObserver needs to function.

src/blockMonitor/blockProcessor.ts
```diff
@@ -1,27 +1,62 @@
...
+    const block = await provider.getBlock(blockNumberOrHash);
+    return {
+        hash: block.hash,
+        number: block.number,
+        parentHash: block.parentHash,
+        transactions: block.transactions
+    };
+};
+
+export const defaultBlockFactory = (provider: ethers.providers.Provider) => async (
```
yahgwai: I would just call this blockFactory

src/blockMonitor/blockProcessor.ts
```diff
@@ -1,27 +1,62 @@
 import { ethers } from "ethers";
 import { StartStopService, ApplicationError } from "../dataEntities";
 import { ReadOnlyBlockCache, BlockCache } from "./blockCache";
-import { IBlockStub } from "./blockStub";
+import { IBlockStub } from "../dataEntities";
+import { Block, HasTxHashes } from "../dataEntities/block";
+
+type BlockFactory<T> = (provider: ethers.providers.Provider) => (blockNumberOrHash: number | string) => Promise<T>;
+
+export const minimalBlockFactory = (provider: ethers.providers.Provider) => async (
```
yahgwai: I would call this blockStubAndTxFactory (even though that's not quite true). Since minimal isn't quite true either,

yahgwai: Both of these factories should have some documentation about what they return

src/dataEntities/block.ts
bigspider: Changed to "Logs" and "Transactions".

src/blockMonitor/blockProcessor.ts
bigspider: Removed, added a comment for future notice.

src/blockMonitor/blockTimeoutDetector.ts
bigspider: Changed to IBlockStub.

src/blockMonitor/confirmationObserver.ts
bigspider: Done.

src/blockMonitor/blockProcessor.ts
bigspider: Done.

src/blockMonitor/blockProcessor.ts
bigspider: Done.

PR #199, opened 2019-07-01
174 responder 2

yahgwai:
Added the tests for multiresponder

### DIFF
--- a/src/responder/gasQueue.ts
+++ b/src/responder/gasQueue.ts
@@ -61,7 +61,7 @@ export class GasQueueItem {
     ) {
         if (currentGasPrice.lt(idealGasPrice)) {
             throw new ArgumentError(
-                "Current gas price must be greater than ideal gas price",
+                "Current gas price cannot be less than ideal gas price",
                 currentGasPrice,
                 idealGasPrice
             );
--- a/src/responder/gasQueue.ts
+++ b/src/responder/gasQueue.ts
@@ -1,5 +1,5 @@
 import { IEthereumResponseData, ArgumentError } from "../dataEntities";
-import { BigNumber, formatUnits } from "ethers/utils";
+import { BigNumber } from "ethers/utils";
 import { ethers } from "ethers";
 
 export class PisaTransactionIdentifier {
@@ -241,9 +241,6 @@ export class GasQueue {
         if (this.depthReached()) {
             throw new ArgumentError(`Cannot add item. Max queue depth reached.`, this.maxQueueDepth);
         }
-        if (this.queueItems.find(q => q.request.identifier.equals(request.identifier))) {
-            throw new ArgumentError("Cannot add item. Item already exists in queue.", request.identifier);
-        }
 
         // starting from the highest gas price, look for the first tx
         // with ideal gas price less than the supplied one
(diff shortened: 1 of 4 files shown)


src/responder/gasQueue.ts
```diff
@@ -288,20 +285,4 @@ export class GasQueue {
...
-    }
-
-    /**
-     * Returns true if a queue item with this identifier exists in this queue
-     * @param identifier
-     */
-    public contains(identifier: PisaTransactionIdentifier) {
-        return this.queueItems.findIndex(i => i.request.identifier.equals(identifier)) !== -1;
-    }
 }
```
bigspider: The error message in the constructor of GasQueueItem says `"Current gas price must be greater than ideal gas price"` but it should say `"Current gas price must not be smaller than ideal gas price"`

test/src/responder/multiResponder.test.ts
```diff
@@ -0,0 +1,277 @@
...
+            errorGasPriceEstimator,
+            transactionTracker,
+            maxConcurrentResponses,
+            replacementRate
+        );
+
+        await responder.startResponse(appointmentId, responseData);
+
+        verify(transactionTrackerMock.addTx(anything(), anything())).never();
+    });
```
bigspider: How does the caller of `startResponse` know that the response was not started successfully? Who is responsible for retrying?
For the dedicated responder, events are emitted and the ResponderManager could react accordingly; what is the plan here?

src/responder/gasQueue.ts
yahgwai: Done.

test/src/responder/multiResponder.test.ts
```diff
@@ -0,0 +1,277 @@
...
+        );
+        await responder.startResponse(appointmentId, responseData);
+        const queueBefore = responder.queue;
+        const item = responder.queue.queueItems[0];
+        await responder.txMined(item.request.identifier, item.nonce + 1);
+
+        expect(responder.queue).to.equal(queueBefore);
+        verify(transactionTrackerMock.addTx(anything(), anything())).once();
+    });
+});
```
bigspider: Perhaps add one or two tests for TransactionTracker.

test/src/responder/multiResponder.test.ts
yahgwai: The plan in the long run will be to throw exceptions and handle these at the responder manager level, where we can retry or ask call other responders. For now I dont want errors to propogate to the watcher, so I but a large try/catch block here, but I could put one in the responder manager instead if you think that's better? My long term plan is to get this working in the anchor state paradigm, and then re-assess how these classes will interact.

test/src/responder/multiResponder.test.ts
yahgwai: My plan is to remove the transaction tracker completley and replace it with an anchor state reducer

test/src/responder/multiResponder.test.ts
bigspider: Fair enough, throwing errors in the Responders and handling them at the level of the ResponderManager seems appropriate to me.

PR #205, opened 2019-07-09
Various improvements

bigspider:
- Added some documentation
- Refactored Watcher to remove StandardMappedComponent, and simplified the predicates
- Modified code of the BlockchainMachine to not assume that the state is computed if a block is in the BlockCache (false on startup)
- Slight simplification in BlockProcessor.

### DIFF
--- a/src/blockMonitor/blockchainMachine.ts
+++ b/src/blockMonitor/blockchainMachine.ts
@@ -54,23 +54,26 @@ export class BlockchainMachine<TBlock extends IBlockStub> extends StartStopServi
         for (const { component, states } of this.componentsAndStates) {
             let state: object | null = null;
             for (const block of ancestorsToAdd) {
-                const parentBlock = this.blockProcessor.blockCache.getBlockStub(block.parentHash);
+                let prevAnchorState: object | undefined;
+                if (this.blockProcessor.blockCache.hasBlock(block.parentHash)) {
+                    const parentBlock = this.blockProcessor.blockCache.getBlockStub(block.parentHash);
+                    prevAnchorState = states.get(parentBlock) || component.reducer.getInitialState(parentBlock);
 
-                // the previous state is the state of the parent block if available, or the initial state otherwise
-                const prevAnchorState = parentBlock
-                    ? states.get(parentBlock)!
-                    : component.reducer.getInitialState(block);
-
-                state = component.reducer.reduce(prevAnchorState, block);
+                    state = component.reducer.reduce(prevAnchorState, block);
+                } else {
+                    state = component.reducer.getInitialState(block);
+                }
 
                 states.set(block, state);
             }
 
             if (state && prevHead) {
-                const prevState = prevHead && states.get(prevHead)!;
-                // TODO:198: should we (deeply) compare old state and new state and only emit if different?
-                // Probably not, it might be expensive/inefficient depending on what is in TState
-                component.handleNewStateEvent(prevHead, prevState, head, state);
+                const prevState = prevHead && states.get(prevHead);
+                if (prevState != null) {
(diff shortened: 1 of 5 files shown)


src/watcher/watcher.ts
```diff
@@ -64,12 +64,26 @@ class AppointmentStateReducer implements StateReducer<WatcherAppointmentState, B
     }
 }
 
+/**
+ * Returns true if and only if the given predicate is true on `(state, block)` and false on `(oldState, oldBlock)`.
+ * TODO:198: could simplify this by removing the block parameter, and changing anchor states accordingly
+ */
+function predicateBecameTrue<TState, TBlock>(
```
yahgwai: You could put this on the Component<T> class

bigspider: Doesn't really work for MappedComponents, though, as the predicate is on single items, rather than on the whole state

bigspider: Perhaps I can just export it in component.ts as a utility function?

yahgwai: Do we have a mappedcomponent, I thought you'd removed that?

bigspider: I removed the StandardMappedComponent, that also implemented handleNewStateEvents as a list of condition/action.
The MappedComponent just computes a reducer by mapping a uniform collection of object (string-indexed) to reducers for each object.

yahgwai: Hmm, still not sure what you mean, I cant see a MappedComponent class on your branch.

bigspider: Oh, right, we only have the MappedStateReducer. But still, Watcher extends Watcher extends Component<MappedState<WatcherAppointmentState>, Block>, while the preedicate is on WatcherAppointmentState only.

yahgwai: OK, I guess the next best place to put it is on the anchor state itself

yahgwai: In fact, we can create methods that match the predicates in the handleStateEvent function. And just call these directly on the state.

src/blockMonitor/blockchainMachine.ts
```diff
@@ -54,23 +54,26 @@ export class BlockchainMachine<TBlock extends IBlockStub> extends StartStopServi
...
 
                 states.set(block, state);
             }
 
             if (state && prevHead) {
-                const prevState = prevHead && states.get(prevHead)!;
-                // TODO:198: should we (deeply) compare old state and new state and only emit if different?
-                // Probably not, it might be expensive/inefficient depending on what is in TState
-                component.handleNewStateEvent(prevHead, prevState, head, state);
+                const prevState = prevHead && states.get(prevHead);
```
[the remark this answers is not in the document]
bigspider: Done in 1d9fc20.

src/blockMonitor/blockchainMachine.ts
```diff
@@ -54,23 +54,26 @@ export class BlockchainMachine<TBlock extends IBlockStub> extends StartStopServi
...
+                if (prevState != null) {
```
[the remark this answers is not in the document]
bigspider: Done in 1d9fc20.

src/blockMonitor/blockchainMachine.ts
```diff
@@ -54,23 +54,26 @@ export class BlockchainMachine<TBlock extends IBlockStub> extends StartStopServi
         for (const { component, states } of this.componentsAndStates) {
             let state: object | null = null;
             for (const block of ancestorsToAdd) {
-                const parentBlock = this.blockProcessor.blockCache.getBlockStub(block.parentHash);
+                let prevAnchorState: object | undefined;
```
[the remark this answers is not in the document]
bigspider: Done in 1d9fc20.

PR #213, opened 2019-07-12
198 responder tests

yahgwai:
Added tests for all new responder classes, components and reducers
Minor refactoring the mapped component


src/responder/component.ts
```diff
@@ -0,0 +1,182 @@
...
+        block: ResponderBlock,
+        identifier: PisaTransactionIdentifier
+    ): { blockNumber: number; nonce: number } | null {
+        for (const tx of block.transactions) {
+            // a contract creation - cant be of interest
+            if (!tx.to) continue;
+
+            // look for matching transactions
+            const txIdentifier = new PisaTransactionIdentifier(tx.chainId, tx.data, tx.to, tx.value, tx.gasLimit);
+            if (txIdentifier.equals(identifier) && tx.from.toLocaleLowerCase() === this.address.toLocaleLowerCase()) {
```
bigspider: I think toLowerCase would be better here, as the locale doesn't matter.

src/blockMonitor/blockchainMachine.ts
```diff
@@ -73,7 +73,7 @@ export class BlockchainMachine<TBlock extends IBlockStub> extends StartStopServi
                 if (prevState) {
                     // TODO:198: should we (deeply) compare old state and new state and only emit if different?
                     // Probably not, it might be expensive/inefficient depending on what is in TState
-                    component.handleNewStateEvent(prevState, state);
+                    component.detectChanges(prevState, state);
```
bigspider: "Detect" seems to suggest a monitoring role rather than an active role. Perhaps just "handleChanges"?

src/responder/gasQueue.ts
```diff
@@ -286,14 +287,55 @@ export class GasQueue {
...
+        }
+
+        // all items ordered by nonce are now also ordered by gas price
+        return new GasQueue(allItemsOrderedByNonce, this.emptyNonce, this.replacementRate, this.maxQueueDepth);
+    }
 
-        throw new ApplicationError("Not implemented");
+    public getItem(request: GasQueueItemRequest) {
+        const queueItem = this.queueItems.find(q => q.request === request);
     }
```
bigspider: I guess you meant to return `queueItem`, but this method is not used anyway

src/responder/multiResponder.ts
```diff
@@ -370,35 +192,28 @@ export class MultiResponder extends StartStopService {
...
-            // TODO:198: add back the contains and difference functions
-            if (this.mQueue.queueItems.findIndex(i => i.request.identifier.equals(item.identifier)) === -1) {
-                missingQueueItems.push(item);
-            }
-        }
-
+        const missingQueueItems = appointmentIdsStillPending
+            .map(a => this.respondedTransactions.get(a))
+            .map(a => {
+                if (!a) throw new ArgumentError("No record of appointment in responder.", a);
```
bigspider: perhaps give some meaningful names to the mapped variable, or at least two different ones to avoid having `a` with two different meanings in nearby rows.

src/blockMonitor/component.ts
```diff
@@ -35,20 +39,23 @@ export class MappedStateReducer<TState extends object, TBlock extends IBlockStub
      */
```
bigspider: Missing documentation for new param `reducer`

test/src/blockMonitor/component.test.ts
```diff
@@ -102,3 +157,21 @@ describe("MappedStateReducer", () => {
...
+        const anchorState = reducer.getInitialState(blocks[0]);
+
+        expect(anchorState.blockNumber).to.equal(blocks[0].number);
+    });
+
+    it("reduce sets current block number", () => {
+        const reducer = new BlockNumberReducer();
+
+        const prevAnchorState = reducer.getInitialState(blocks[0]);
+        const nextAnchorState = reducer.reduce(prevAnchorState, blocks[2]);
```
bigspider: It doesn't matter much here, but I think it's better to explicitly set `prevAnchorState = { ... }` instead of using `getInitialState`.

test/src/responder/component.test.ts
```diff
@@ -0,0 +1,304 @@
...
+): MinedResponseState => {
+    const identifier = new PisaTransactionIdentifier(1, data, "to", new BigNumber(0), new BigNumber(200));
+    return {
+        appointmentId,
+        blockMined,
+        identifier,
+        kind: ResponderStateKind.Mined,
+        nonce
+    };
+};
```
bigspider: Not important, but I prefer to call these type of functions `makeX()` instead of `X()`.

test/src/responder/gasQueue.test.ts
```diff
@@ -313,4 +321,166 @@ describe("GasQueue", () => {
...
+
+        const q = new GasQueue(items, 4, 15, 5);
+
+        const contains1 = q.contains(id1);
+        const contains2 = q.contains(id2);
+        const contains3 = q.contains(id3);
+
+        expect(contains1).to.be.true;
+        expect(contains2).to.be.true;
+        expect(contains3).to.be.true;
```
bigspider: Could merge these 3 lines with the previous 3.

src/responder/gasQueue.ts
```diff
@@ -286,14 +287,55 @@ export class GasQueue {
...
+    public contains(identifier: PisaTransactionIdentifier): boolean {
+        return this.queueItems.findIndex(i => i.request.identifier.equals(identifier)) !== -1;
+    }
+
+    
+    /**
+     * Re-add some items that have lower nonces than any in the current queue
+     * @param lowerNonceItems 
+     */
+    public unlock(lowerNonceItems: GasQueueItem[]): GasQueue {
```
bigspider: The name `unlock` is rather confusing, as it doesn't match what the code does and it's not clear in what sense items are "locked".

test/src/responder/multiResponder.test.ts
```diff
@@ -301,12 +310,137 @@ describe("MultiResponder", () => {
...
+            signer,
+            decreasingGasPriceEstimator,
+            maxConcurrentResponses,
+            replacementRate
+        );
+        await responder.start();
+
+        expect(responder.reEnqueueMissingItems([appointmentId])).to.eventually.be.rejectedWith(ArgumentError);
+
+        await responder.stop();
```
bigspider: Shouldn't we wait for reEnqueueMissingItems to complete, before calling `responder.stop()`?

src/responder/component.ts
yahgwai: Yep, locale was a mistake.

src/blockMonitor/blockchainMachine.ts
yahgwai: Changed

src/responder/gasQueue.ts
yahgwai: Removed

src/responder/multiResponder.ts
yahgwai: Updated

test/src/responder/gasQueue.test.ts
yahgwai: Done

src/responder/gasQueue.ts
yahgwai: Changed to prepend

test/src/responder/multiResponder.test.ts
yahgwai: Updated

Issue #121, opened 2019-05-02
Start stop services should not allow public methods to be called if they haven't been started.


yahgwai:
At the moment we have a number of start stop services eg. watcher, gc. These services require a .start() function to be called to function correctly.

At the moment we do nothing to prevent this. One way we could do this is to check the value of the started property at the start of every public method, if it is true then we can continue, otherwise throw error. A better way would be to have this done automatically, perhaps using a proxy classm to augment all public functions and properties with this check.

PR #224, opened 2019-07-22
#121 Error when startstopservice not started 

morkeltry:
#121 
`startStopService`'s constructor returns a proxy, on which:
`get` is transparent when getting reference to `startInternal`, `stopInternal`, `start` and `stop` methods; 
or also if the `startStopService` instance is started (as per internal property `mStarted`
or also if a flag `suppressNotStartedError` (currently number) is set and not unset by one of those methods;
(and thus if a different method called below one of those methods)

`construct` is differently intercepted, based on examining the stack trace of a `new Error` (not thrown)

Can throw a 'Service not started' instance of ApplicationError.
More detailed logging in the error is available, done manually depending on property `logVerbosityOfNotStartedError` of the class.

Test added in `startStop.test.ts` expecting `ApplicationError` with the string beginning  `'Service not started'` when the empty `emptyTestMethod` is called on an instance without previously calling `start`.


morkeltry:
Superceded : https://github.com/PISAresearch/pisa/pull/281/files

Issue #218, opened 2019-07-16
Remove the garbage collector.


yahgwai:
We use the garbage collector to check for expired appointments and remove them from the db. We now have a framework for processing new blocks, and we can use this to check if an appointment has expired instead of using a specific garbage collector.

In the watcher appointment reducer we set new anchor state based on a newly received block. For each appointment we should check if the new block has a number >= appointment.endBlock + someThreshold. The threshold is necessary in case we have reorgs and need to recover the state. If the predicate is true, we should set a status to EXPIRED. In the watcher component we can then check for EXPIRED appointments and remove them the db.

PR #231, opened 2019-07-25
Remove gc and update Watcher

irisradoi:
Related to https://github.com/PISAresearch/pisa/issues/218


package.json
```diff
@@ -54,7 +54,7 @@
         "express": "^4.16.4",
         "express-http-context": "^1.2.0",
         "express-rate-limit": "^3.5.0",
-        "ganache-core": "^2.5.6",
```
yahgwai: You should revert this change. We try to keep packages the same, then upgrade them explicitly as part of an issue.

You can use `npm ci` to install the exact versions that are in the package-lock.json

src/watcher/watcher.ts
```diff
@@ -121,6 +121,15 @@ export class Watcher extends Component<WatcherAnchorState, IBlockStub & Logs> {
         appointmentState.state === WatcherAppointmentState.OBSERVED &&
         state.blockNumber - appointmentState.blockObserved + 1 >= this.confirmationsBeforeRemoval;
 
+    private shouldRemoveExpiredAppointment = (
```
yahgwai: Can you change the name of `shouldRemoveAppointment` to `shouldRemoveObservedAppointment` to avoid ambiguity.

irisradoi: Done

package.json
irisradoi: Done

PR #236, opened 2019-07-26
173 new appointment

yahgwai:


src/dataEntities/appointment.ts
```diff
@@ -1,149 +1,328 @@
...
+            eventArgs: appointment.eventArgs,
+            postCondition: appointment.postCondition,
+            paymentHash: appointment.paymentHash
+        };
     }
 
     /**
-     * A combination of the state locator and the nonce for this state update
+     * The hash provided for free the access
      */
```
bigspider: Add a more descriptive comment, perhaps.

src/dataEntities/appointment.ts
```diff
@@ -1,149 +1,328 @@
...
-            ...this,
-            startBlock: this.startBlock,
-            endBlock: this.endBlock
+            address: this.contractAddress,
+            topics
         };
     }
+    /**
+     * Th packed representation for the
+     */
```
bigspider: Incomplete comment.

src/dataEntities/appointment.ts
```diff
@@ -1,149 +1,328 @@
...
+                this.refund,
+                this.gas,
+                this.mode,
+                ethers.utils.toUtf8Bytes(this.eventABI),
+                this.eventArgs,
+                this.postCondition,
+                this.paymentHash
+            ]
+        );
+    }
```
bigspider: If ethers.js doesn't have a way of encoding lists of arguments as arrays of
```
{
 type: "<>",
 value: <>
}
```
or something similar, perhaps we should write one!

src/dataEntities/errors.ts
```diff
@@ -1,3 +1,5 @@
+import { JsonRpcProvider } from "ethers/providers";
+
```
bigspider: Unused import

src/dataEntities/appointment.ts
yahgwai: Added a helper function called "groupTuples"

src/dataEntities/appointment.ts
yahgwai: Corrected

src/dataEntities/appointment.ts
yahgwai: Added

src/dataEntities/errors.ts
yahgwai: Removed

PR #238, opened 2019-07-29
Added relay functionality with end to end test

yahgwai:
Also updated the mocks for the multiResponder tests as I noticed that some where throwing errors since we were using ganache but not passing correct transaction data.


test/src/serviceEndToEnd.test.ts
```diff
@@ -172,6 +177,31 @@ describe("Service end-to-end", () => {
         }
     }).timeout(3000);
 
+    it("create channel, relay trigger disupte", async () => {
```
bigspider: typo "disupte"

PR #240, opened 2019-07-29
BlockProcessor and BlockCache robustness and refactoring (WIP)

bigspider:
Not quite ready to merge, but creating a PR anyway to get feedback. Description of the changes:

1) Modified the BlockCache to also allow adding blocks before adding their parent. Those blocks are considered "pending" and kept in a separate structure. `getBlock` will return a known block regardless if it's pending, while `hasBlock` allow to specify if pending blocks should be included or not. The addition of pending blocks allows to simplify the BlockProcessor's logic, at the cost of some more complexity in the BlockCache, as I added some non-trivial logic to make "pending" block become "complete" as soon as possible, while staying close to computationally optimal.
2) The BlockProcessor's logic is a bit different than before, in order to tolerate failures (e.g. from getBlock unexpectedly returning null). Right now, the behavior I implemented is a silent failure (which means, it will try again at the next new_head event from the provider, but previously downloaded blocks are added to the cache (as pending), so they would't be downloaded again.
3) I also changed the behavior of the NEW_BLOCK event of the BlockProcessor, which is now re-emitting blocks that were already emitted in case of switching forks multiple times. It would be easy to change this back, and I did not yet wrap my head what's best.

All in all, I'm quite happy with the test coverage of the BlockCache, while I trust the BlockProcessor work a bit less; it's a bit harder to think of good tests here, though.

### DIFF
--- a/src/blockMonitor/blockCache.ts
+++ b/src/blockMonitor/blockCache.ts
@@ -21,20 +21,13 @@ export interface ReadOnlyBlockCache<TBlock extends IBlockStub> {
  * Utility class to store and query info on full blocks up to a given maximum depth `maxDepth`, compared to the current
  * maximum height ever seen.
  * It prunes all the blocks at depth bigger than `maxDepth`, or with height smaller than the first block that was added.
- * It does not allow to add blocks without adding their parent first, except if they are at depth `maxDepth`.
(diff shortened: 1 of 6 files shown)


bigspider:
Closing, will reopen from another branch.

PR #241, opened 2019-07-30
'IT' overload created for easier test refactoring

irisradoi:
Added library fnIt that can be used to design tests that will automatically update the name of the function tested in the test description.


test/src/blockMonitor/blockCache.test.ts
```diff
@@ -1,7 +1,8 @@
 import "mocha";
 import { expect } from "chai";
-import { BlockCache, getConfirmations } from "../../../src/blockMonitor";
-import { ArgumentError, IBlockStub, TransactionHashes, ApplicationError } from "../../../src/dataEntities";
+import { BlockCache, getConfirmations, ReadOnlyBlockCache } from "../../../src/blockMonitor";
+import { ArgumentError, IBlockStub, TransactionHashes, ApplicationError, Block } from "../../../src/dataEntities";
```
yahgwai: This Block import isnt used. Neither is the ReadOnlyBlockCache one.

utils/fnIt.ts
```diff
@@ -0,0 +1,13 @@
+import { spy } from "ts-mockito";
+
```
yahgwai: This utils folder is at the top level. I think you should put it in the /test/ folder

utils/fnIt.ts
```diff
@@ -0,0 +1,13 @@
...
+const getName = <T>(fn: (arg: T) => (...args: any[]) => any ) => {
+    // create an empty instance
+    const dummyInstance = {} as T;
+    // spy it to mark up some properties
+    const dummySpy: T = spy(dummyInstance);
+    // used this 'typed' spy to extract the name
+    return fn(dummySpy).name;
+}
+export function fnIt<T>(fn: (t:T) => (...args:any[]) => any, message : string, test: () => void){
```
yahgwai: Exported functions, and public methods/properties on classes, should have a comment to describe what they are and how they're used.
Try not to describe the what the function does internally (mocks etc), and try not to repeat anything that is in the function signature/name. More try to describe what the motivation is for the function, and how it can be used.

test/src/watcher/watcher.test.ts
```diff
@@ -377,7 +377,7 @@ describe("Watcher", () => {
         verify(mockedStore.removeById(appointment.id)).once();
     });
 
-    it("handleChanges dooes not call removeById before an appointment has expired for long enough", async () => {
+    fnIt<Watcher>(w => w.handleChanges, "dooes not call removeById before an appointment has expired for long enough", async() => {
```
yahgwai: Not your typo - it was there already - but since you're editing this line could fix the `dooes`

test/src/watcher/watcher.test.ts
```diff
@@ -179,36 +180,35 @@ describe("Watcher", () => {
...
-            {
-                items: makeMap(appointment.id, {
-                    state: WatcherAppointmentState.OBSERVED,
-                    blockObserved: 2
-                }),
-                blockNumber: 2 + CONFIRMATIONS_BEFORE_RESPONSE - 1
-            }
-        );
-
+        responder,
```
yahgwai: This block has been back tabbed for some reason.

utils/fnIt.ts
yahgwai: Since this is the only function in this file (and it has the same name as the file), it's probably appropriate to do a `default` export. This allows for slightly cleaner imports in other files. Not essential though, just pointing it out so you know `default` is there.

test/src/responder/gasQueue.test.ts
```diff
@@ -445,7 +446,7 @@ describe("GasQueue", () => {
         expect(uQ.difference(q)).to.deep.equal([]);
     })
 
-    it("unlock does throws error for missing nonces", () => {
+    it("unlock does throw error for missing nonces", () => {
```
yahgwai: This should be `prepend` - the function used to be called `unlock` but refactoring failed here.

test/src/responder/gasPriceEstimator.test.ts
```diff
@@ -127,8 +128,7 @@ describe("GasPriceEstimator", () => {
             startBlock: 7
         });
     };
-
-    it("estimate", async () => {
+        fnIt<GasPriceEstimator>(e => e.estimate, "", async () => {
```
yahgwai: An extra tab here. Salvatore and I use an vs code plugin called "Prettier", it stops use making different formatting changes from each other. You should add it then set the following options in your vs code settings:

```
"prettier.tabWidth": 4,
    "prettier.printWidth": 120,
    "[typescript]": {
        "editor.defaultFormatter": "esbenp.prettier-vscode"
    },
    "[json]": {
        "editor.defaultFormatter": "esbenp.prettier-vscode"
    },
```

test/src/blockMonitor/component.test.ts
```diff
@@ -143,7 +144,7 @@ describe("MappedStateReducer", () => {
         expect(reducedState.items).to.deep.equal(expectedMap);
     });
 
-    it("reduce calls getInitialState if a new object id is added to the collection", () => {
+    fnIt<MappedStateReducer<any,any,any,any>>(m => m.reduce, "calls getInitialState if a new object id is added to the collection", () => {
```
yahgwai: Wow, this is interesting, I hadn't thought about generics. I think using the <any> is fine, maybe you should point that out in the documentation you write on `fnIt`

test/src/blockMonitor/blockCache.test.ts
```diff
@@ -2,6 +2,7 @@ import "mocha";
 import { expect } from "chai";
 import { BlockCache, getConfirmations } from "../../../src/blockMonitor";
 import { ArgumentError, IBlockStub, TransactionHashes, ApplicationError } from "../../../src/dataEntities";
+import fnIt from "../../../test/utils/fnIt";
```
yahgwai: You can remove a `../test` from this - since we're already in the test folder

test/utils/fnIt.ts
```diff
@@ -0,0 +1,16 @@
...
+const getName = <T>(fn: (arg: T) => (...args: any[]) => any) => {
+    // create an empty instance
+    const dummyInstance = {} as T;
+    // spy it to mark up some properties
+    const dummySpy: T = spy(dummyInstance);
+    // used this 'typed' spy to extract the name
+    return fn(dummySpy).name;
+};
+
+/* fnIt can be used to retrieve function names rather than writing them manually. This will be useful when refactoring
```
yahgwai: You put the comment in the js doc style:
```
/**
 *
 */
```
If you do that it'll show up when someone hovers over `fnIt` in any file. Also I like the second part of the comment, but I dont think the first part is true: getName can be used to retrieve function names, fnIt cant. I would write something like:
`A test fixture for a specific function. The supplied function will be used to name the test. This will be useful when since refactoring there is no need to go back an change all the tests manually in case a function name has been changed. Types of T that are generic can be declared with 'any' eg. fnIt<BlockCache<any>>(b => b.setHead, ...)`

PR #254, opened 2019-08-05
239 mocks stubbing error

irisradoi:


test/utils/throwingInstance.ts
```diff
@@ -0,0 +1,24 @@
...
+                return Reflect.get(target, prop);
+            } else {
+                //We use this to print the error. Throwing an error will not always be called as some other error(indirect consquence of not stubbing) might catch and not log the error
+                console.log(prop + " has not been stubbed. Use when() to stub this method before calling it.");
+            }
+        }
+    };
+
+    stubbedMethods = Object.keys((target as any)["tsMockitoInstance"]["mocker"]["methodStubCollections"]);
+    let p = new Proxy(instance(target), handler);
```
yahgwai: no need to assign a variable here - just return.

test/utils/throwingInstance.ts
```diff
@@ -0,0 +1,24 @@
+import { instance } from "ts-mockito";
+
+/**
+ * This will be useful for detecting when certain methods have not been stubbed.
+ * @param target will take the mocked object
+ */
+export default function throwingInstance(target: any) {
```
yahgwai: This function returns `any`. In doing so we lose the type annotations when we pass an object through throwingInstance. This function shouldn't expect any - it should expect a generic type T - which you should use wherever possibly in this function, including the return type.

test/utils/throwingInstance.ts
```diff
@@ -0,0 +1,24 @@
+import { instance } from "ts-mockito";
+
+/**
+ * This will be useful for detecting when certain methods have not been stubbed.
```
yahgwai: Comments should mostly describe a high level of what the function/object does. In this case I would say something like: `Creates an instance of a mocked object. Calls to methods or properties that have not been stubbed will throw errors.`

test/src/responder/multiResponder.test.ts
```diff
@@ -46,12 +47,12 @@ describe("MultiResponder", () => {
...
         when(providerMock.getNetwork()).thenResolve({ chainId: 1, name: "test" });
         when(providerMock.getTransactionCount("address", "pending")).thenResolve(1);
-        const provider = instance(providerMock);
+        const provider = throwingInstance(providerMock);
 
         const signerMock = mock(ethers.providers.JsonRpcSigner);
         when(signerMock.getAddress()).thenResolve("address");
         when(signerMock.provider).thenReturn(provider);
-        signer = instance(signerMock);
+        signer = throwingInstance(signerMock);
```
yahgwai: I think you can add
`when(signerMock.sendTransaction(anything())).thenResolve()`

to mock the sendTransaction

src/responder/multiResponder.ts
```diff
@@ -99,13 +99,15 @@ export class MultiResponder extends StartStopService {
...
             );
+
             const idealGas = await this.gasEstimator.estimate(appointment);
             const request = new GasQueueItemRequest(txIdentifier, idealGas, appointment);
             logger.info(request, `Enqueueing request for ${appointment.id}.`);
 
             // add the queue item to the queue, since the queue is ordered this may mean
             // that we need to replace some transactions on the network. Find those and
             // broadcast them
+            //PROBLEM HERE
```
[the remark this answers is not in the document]
irisradoi: i used this during debugging, thanks for noticing

package.json
```diff
@@ -62,6 +62,7 @@
         "request": "^2.88.0",
         "rwlock": "^5.0.0",
         "swagger-jsdoc": "^3.3.0",
+        "truffle-assertions": "^0.9.1",
```
yahgwai: This seems to have been added by mistake.

test/utils/throwingInstance.ts
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Creates an instance of a mocked object. Calls to methods or properties that have not been stubbed will throw errors
+ */
+export default function throwingInstance<T>(target: T) {
+    const stubbedMethods: Array<string> = Object.keys(
+        (target as any)["tsMockitoInstance"]["mocker"]["methodStubCollections"]
+    );
+
+    const handler = {
+        get: function(target: any, prop: string, receiver: any) {
```
yahgwai: The target here can be T, as it turns out you're not using any of the typing in this function. But it's still good practice

test/utils/throwingInstance.ts
```diff
@@ -0,0 +1,23 @@
+import { instance } from "ts-mockito";
+
+/**
+ * Creates an instance of a mocked object. Calls to methods or properties that have not been stubbed will throw errors
+ */
+export default function throwingInstance<T>(target: T) {
```
yahgwai: I forgot to mention this last time, but it's good to call the T variable something a bit more meaningful. Maybe TMock would be good in this case.

test/src/blockMonitor/blockProcessor.test.ts
```diff
@@ -2,10 +2,11 @@ import "mocha";
 import * as chai from "chai";
 import chaiAsPromised from "chai-as-promised";
 import { ethers } from "ethers";
-import { mock, when, instance, anything } from "ts-mockito";
+import { mock, when, anything, capture, anyNumber } from "ts-mockito";
```
yahgwai: Unused import 'capture'

package-lock.json
```diff
@@ -15813,6 +15812,15 @@
       "integrity": "sha1-yy4SAwZ+DI3h9hQJS5/kVwTqYAM=",
       "dev": true
     },
+    "truffle-assertions": {
```
yahgwai: The package-lock.json still contains some unexpected changes.

PR #258, opened 2019-08-06
Create .gitattributes so that Solidity renders in Github

gakonst:
https://medium.com/@danielque/psa-how-to-fix-githubs-syntax-highlighting-for-solidity-4e9867c540b6


yahgwai:
This looks great, however it seems to work on all our .sol files except PisaHash.sol. When I view this file I dont see syntax highlighting: https://github.com/gakonst/pisa/blob/patch-1/sol/contracts/PISAHash.sol. Do you? Any idea why?

PR #262, opened 2019-08-08
261 data persistence for responder

yahgwai:


src/responder/store.ts
```diff
@@ -0,0 +1,115 @@
...
+     * Update the queue. Returns a new transactions that need to be issued as result of the update.
+     * @param queue
+     */
+    public async updateQueue(queue: GasQueue) {
+        // const replacedQueue = this.zQueue.add(request);
+        const difference = queue.difference(this.mQueue);
+        this.mQueue = queue;
+
+        // update these transactions locally and in the db
+        const differencyById = new Map<string, GasQueueItem>();
```
bigspider: typo `differencyById` => `differenceById`

src/responder/gasQueue.ts
```diff
@@ -94,7 +167,32 @@ export class GasQueueItem {
...
+
+    public static serialise(queue: GasQueue): GasQueueSerialisation {
+        return {
+            queueItems: queue.queueItems.map(i => GasQueueItem.serialise(i)),
+            emptyNonce: queue.emptyNonce,
+            maxQueueDepth: queue.maxQueueDepth,
+            replacementRate: queue.replacementRate
+        };
+    }
+
```
bigspider: I'm not sure I like this approach of having a `-Serialisation` interface for each class that we want to serialise. While it adds some type safety, it does not enforce that all the serializable objects have the same "serialise"/"deserialise" methods; and the type safety is in a place where is probably not really useful (after you serialize an object, you are usually about to write it to DB after all).

I was thinking of a simple common interface like this:

```
interface Serialisable {
    serialise(): object;
}
```

the serialise() object is then called on the item itself rather than being a static class method.
A downside is that static methods cannot be specified in the interface, so still there is no way of forcing subclasses to implement a `deserialise` method.

Another concern is that the deserialisation is currently not checking for any kind of error (e.g. missing or wrong fields, or unexpected fields); but perhaps this should be a separate ticket.

test/src/responder/gasQueue.test.ts
```diff
@@ -482,4 +482,18 @@ describe("GasQueue", () => {
...
+    fnIt<GasQueue>(() => GasQueue.serialise, "correctly deserialises and serialises", () => {
+        const item1 = createGasQueueItem(1, 1, new BigNumber(90), new BigNumber(90), createIdentifier("data1", "to1"));
+        const item2 = createGasQueueItem(2, 2, new BigNumber(50), new BigNumber(60), createIdentifier("data2", "to2"));        
+        const item3 = createGasQueueItem(3, 3, new BigNumber(30), new BigNumber(40), createIdentifier("data3", "to3"));        
+
+
+        const queue = new GasQueue([item1, item2, item3], 4, 12, 10)
+
+        const serialised = GasQueue.serialise(queue);
+        const deserialisd = GasQueue.deserialise(serialised);
```
bigspider: typo `deserialisd` => `deserialised`

src/responder/store.ts
```diff
@@ -0,0 +1,115 @@
...
+    private readonly subDb: LevelUp<EncodingDown<string, any>>;
+    private readonly queueKey: string;
+    /**
+     * A persistent store for responder data.
+     * @param db A backend database for the store
+     * @param responderAddress The address of the responder using this store. Responder public keys can only
+     * be used by one responder at a time.
+     */
+    constructor(db: LevelUp<EncodingDown<string, any>>, responderAddress: string, seedQueue: GasQueue) {
+        //TODO:261: this encoding is actually a requirement
```
bigspider: TODO yet to be addressed?

src/responder/component.ts
```diff
@@ -230,7 +230,7 @@ export class MultiResponderComponent extends Component<ResponderAnchorState, Blo
                     await this.responder.txMined(action.identifier, action.nonce);
                     break;
                 case ResponderActionKind.EndResponse:
-                    this.responder.endResponse(action.appointmentId);
+                    await this.responder.endResponse(action.appointmentId);
                     break;
                 default:
                     throw new ArgumentError("Unrecognised action kind.", action);
```
bigspider: I previously added an `UnreachableCaseError` for this kind of things, if you like to use it; has the advantage of being type safe at compile-time.

test/src/serviceEndToEnd.test.ts
```diff
@@ -238,17 +251,18 @@ describe("Service end-to-end", () => {
...
             json: appRequest
         });
-        expect(request.post(`http://${nextConfig.hostName}:${nextConfig.hostPort}/appointment`, {
-            json: appRequest
-        })).to.eventually.be.rejected;
+        expect(
+            request.post(`http://${nextConfig.hostName}:${nextConfig.hostPort}/appointment`, {
+                json: appRequest
+            })
+        ).to.eventually.be.rejected;
```
bigspider: I don't think this works as expected, as I had this problem before; this will likely pass regardless of the outcome, as the test ends before the promise returns. To use the `to.eventually.` syntax, it should be the last condition of the test and be returned, e.g.:
```
    return expect(something).to.eventually.be.rejected;
```
See https://www.chaijs.com/plugins/chai-as-promised/ for more on that.

I just checked that there are multiple other places in the codebase using this incorrectly, opening a separate ticket.

src/responder/gasQueue.ts
yahgwai: Yeah, `deserialise` is the problem. Since I cant have an interface for `deserialise` I'd rather keep both methods together.

Regarding errors, this deserialise isnt a checking method - we should never use it for data that isnt under our control. We should always be sure of the soundness of the data before calling deserialise - in this case soundness is ensured since we know the data was serialised by the `serialise` method. Later on we may need to add some specific checks (maybe when we make backward breaking changes to the serialise method)

src/responder/store.ts
yahgwai: Updated

test/src/responder/gasQueue.test.ts
yahgwai: Updated

src/responder/store.ts
yahgwai: No, it's been addressed. The sub db needs to explicitly insist on json encoding, rather than relying on the encoding of the passed in db

src/responder/component.ts
yahgwai: Now using UnreacheableCaseError

test/src/serviceEndToEnd.test.ts
yahgwai: I'll fix this as part of #265

src/responder/gasQueue.ts
bigspider: Yeah, my concern about error checking was to protect from
- database corruptions
- bugs in the (de)serialisation code, at least to catch them
- version changes as you mentioned.

Anyway, I agree that it's not the most pressing issue.

PR #264, opened 2019-08-08
230 safe side effects

yahgwai:
Just checked that they can run concurrently - which they can as of #262 Which needs to be merged first.

### DIFF
--- a/src/blockMonitor/component.ts
+++ b/src/blockMonitor/component.ts
@@ -118,7 +118,7 @@ export abstract class Component<
      * @param prevState
      * @param state
      */
-    public abstract async applyAction(actions: TAction): Promise<void>;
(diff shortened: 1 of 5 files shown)


src/blockMonitor/component.ts
```diff
@@ -114,10 +114,11 @@ export abstract class Component<
     constructor(public readonly reducer: StateReducer<TState, TBlock>) {}
     /**
      * Triggers side effects specified by the actions
+     * All side-effect must be thread safe so that they can be applied concurrently
      * @param prevState
      * @param state
      */
-    public abstract async handleChanges(actions: TAction[]): Promise<void>;
+    public abstract async applyAction(actions: TAction): Promise<void>;
```
bigspider: Change `actions` to `action`.

PR #266, opened 2019-08-12
Added new action to check responder balance

irisradoi:
Created new action called `ResponderLowBalance` that will check wether the responder has enough balance. If not the logger will output and error to notify of this issue.

### DIFF
--- a/src/responder/component.ts
+++ b/src/responder/component.ts
@@ -115,7 +115,8 @@ export class ResponderAppointmentReducer implements StateReducer<ResponderAppoin
 export enum ResponderActionKind {
     ReEnqueueMissingItems = 1,
     TxMined = 2,
-    EndResponse = 3
+    EndResponse = 3,
+    ResponderLowBalance = 4
 }
 
 export type ReEnqueueMissingItemsAction = {
@@ -134,17 +135,23 @@ export type EndResponseAction = {
     readonly appointmentId: string;
 };
 
-export type ResponderAction = TxMinedAction | ReEnqueueMissingItemsAction | EndResponseAction;
+export type ResponderLowBalance = {
+    readonly kind: ResponderActionKind.ResponderLowBalance;
+}
+
+
+export type ResponderAction = TxMinedAction | ReEnqueueMissingItemsAction | EndResponseAction | ResponderLowBalance;
 
 /**
  * Handle the state events related to the multiresponder. Knows how to interpret
  * changes in the responder anchor state, and when to fire side effects.
  */
 export class MultiResponderComponent extends Component<ResponderAnchorState, Block, ResponderAction> {
+    
     public constructor(
         private readonly responder: MultiResponder,
         blockCache: ReadOnlyBlockCache<Block>,
-        private readonly confirmationsRequired: number
+        private readonly confirmationsRequired: number,
     ) {
         super(
             new MappedStateReducer(
@@ -202,8 +209,11 @@ export class MultiResponderComponent extends Component<ResponderAnchorState, Blo
                     kind: ResponderActionKind.TxMined,
                     identifier: currentItem.identifier,
                     nonce: currentItem.nonce
+                },{
+                    kind:ResponderActionKind.ResponderLowBalance,
                 });
-            }
+                
+            }   
 
             // after a certain number of confirmations we can stop tracking a transaction
             if (
@@ -232,6 +242,11 @@ export class MultiResponderComponent extends Component<ResponderAnchorState, Blo
             case ResponderActionKind.EndResponse:
                 await this.responder.endResponse(action.appointmentId);
                 break;
+            case ResponderActionKind.ResponderLowBalance:
+                if((await this.responder.signer.provider!.getBalance(this.responder.address)).gt("500000000000000000")){
+                    logger.error("Responder balance is becoming low. Current balance: "+ (await this.responder.signer.provider!.getBalance(this.responder.address)));
+                 }
+                 break;
             default:
                 throw new UnreachableCaseError(action, "Unrecognised responder action kind.");
         }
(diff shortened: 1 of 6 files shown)


src/responder/component.ts
```diff
@@ -115,7 +115,8 @@ export class ResponderAppointmentReducer implements StateReducer<ResponderAppoin
 export enum ResponderActionKind {
     ReEnqueueMissingItems = 1,
     TxMined = 2,
-    EndResponse = 3
+    EndResponse = 3,
+    ResponderLowBalance = 4
```
yahgwai: I think this should be renamed to `CheckResponderBalance`, since when the action is issued we dont know whether the balance is low or not.

src/responder/component.ts
```diff
@@ -232,6 +242,11 @@ export class MultiResponderComponent extends Component<ResponderAnchorState, Blo
             case ResponderActionKind.EndResponse:
                 await this.responder.endResponse(action.appointmentId);
                 break;
+            case ResponderActionKind.ResponderLowBalance:
+                if((await this.responder.signer.provider!.getBalance(this.responder.address)).gt("500000000000000000")){
```
yahgwai: I think you can package this up into a function on the responder called `checkBalance`, which internally checks balance and logs if necessary. From here you would then just call `await responder.checkBalance()`.

yahgwai: The value `500000000000000000` should be a configurable. Set is a constructor argument, low balance warning, document it, and then pass in the `500000000000000000` when a new responder is instantiated.

test/src/responder/component.test.ts
```diff
@@ -7,17 +7,19 @@ import {
...
 import { PisaTransactionIdentifier } from "../../../src/responder/gasQueue";
-import { BigNumber } from "ethers/utils";
+import { BigNumber, bigNumberify } from "ethers/utils";
 import { ResponderBlock, TransactionStub, Block } from "../../../src/dataEntities/block";
 import { expect } from "chai";
 import { MultiResponder } from "../../../src/responder";
 import { mock, anything, when } from "ts-mockito";
 import fnIt from "../../utils/fnIt";
 import throwingInstance from "../../utils/throwingInstance";
+import logger from "../../../src/logger";
```
yahgwai: Unused import

src/responder/component.ts
```diff
@@ -202,8 +209,11 @@ export class MultiResponderComponent extends Component<ResponderAnchorState, Blo
                     kind: ResponderActionKind.TxMined,
                     identifier: currentItem.identifier,
                     nonce: currentItem.nonce
+                },{
```
yahgwai: I think a comment here about why we should check balance after a tx is mined would be appropriate.

src/responder/component.ts
```diff
@@ -134,17 +135,23 @@ export type EndResponseAction = {
     readonly appointmentId: string;
 };
 
-export type ResponderAction = TxMinedAction | ReEnqueueMissingItemsAction | EndResponseAction;
+export type CheckRespnderBalanceAction = {
```
yahgwai: Typo `Respnder`

src/responder/multiResponder.ts
```diff
@@ -33,16 +34,21 @@ export class MultiResponder {
      *   The signer used to sign transaction created by this responder. This responder
      *   requires exclusive use of this signer.
      * @param gasEstimator
+     * @param responderBalanceTreshold
```
yahgwai: Typo `Treshold`

yahgwai: Also, would be good to include the units in the comments - wei.

src/responder/multiResponder.ts
```diff
@@ -33,16 +34,21 @@ export class MultiResponder {
...
         public readonly gasEstimator: GasPriceEstimator,
         private readonly chainId: number,
         store: ResponderStore,
-        public readonly address: string
+        public readonly address: string,
+        public readonly responderBalanceTreshold: number
     ) {
         this.broadcast = this.broadcast.bind(this);
         this.zStore = store;
+        this.balanceTreshold = responderBalanceTreshold;
```
yahgwai: You dont need to set this here, since responderBalanceThreshold is declared `public readonly` in the constructor. https://kendaleiv.com/typescript-constructor-assignment-public-and-private-keywords/

I think the name `balanceThreshold` is good for this property.

src/responder/multiResponder.ts
```diff
@@ -207,4 +213,10 @@ export class MultiResponder {
             logger.error(doh);
         }
     }
+
+    public async checkBalance(){
+        if((await this.signer.provider!.getBalance(this.address)).gt(this.balanceTreshold)){
```
yahgwai: Public functions should have comment explaining what they do.

src/responder/multiResponder.ts
```diff
@@ -207,4 +213,10 @@ export class MultiResponder {
...
+            logger.error("Responder balance is becoming low. Current balance: "+ (await this.signer.provider!.getBalance(this.address)));
```
yahgwai: Calling getBalance twice means you make two network requests. Better to assign the result to a variable instead.

src/responder/multiResponder.ts
```diff
@@ -207,4 +211,14 @@ export class MultiResponder {
             logger.error(doh);
         }
     }
+    /**
+     * Checks to see if the responder balance is lower than the threshold set in the constructor.
+     * If the balance is lower, a warning will be outputted by the logger
+     */
+    public async checkBalance(){
+        const currentBalance = await this.signer.provider!.getBalance(this.address);
+        if(currentBalance.gt(this.balanceThreshold)){
```
yahgwai: Woops, I've just realised that this check is the wrong way round. Should be `.lt`

irisradoi: Thanks for pointing this out, I switched it to test something and forgot to change it back. There would have been a lot of low balance warnings issued lol

PR #270, opened 2019-08-13
Minor logging improvements

yahgwai:
After creating some basic monitoring tools, some gaps in the logging where noticed. These have been added on this branch.


src/dataEntities/appointment.ts
```diff
@@ -217,7 +218,7 @@ export class Appointment {
      */
     public static FreeHash = ethers.utils.keccak256(ethers.utils.toUtf8Bytes("on-the-house")).toLowerCase();
 
-    static parseBigNumber(numberString: string, name: string) {
+    static parseBigNumber(numberString: string, name: string, log: Logger) {
         try {
```
bigspider: `log` parameter is not used.

src/dataEntities/appointment.ts
```diff
@@ -231,31 +232,32 @@ export class Appointment {
     /**
      * Parse an appointment and check property types.
      * @param obj
+     * @param log Logger to be used in case of failures
      */
-    public static parse(obj: any) {
+    public static parse(obj: any, log?: Logger) {
```
bigspider: perhaps a default parameter `log?: Logger = logger` would be simpler, instead of using `(log || logger)` later?

src/service.ts
```diff
@@ -151,13 +156,31 @@ export class PisaService extends StartStopService {
...
+            res.on("finish", () => {
+                const endNano = process.hrtime.bigint();
+                const microDuration =   Number.parseInt((endNano - startNano).toString()) / 1000
+                const logEntry = { req: req, res: res, duration: microDuration };
+
+                if (res.statusCode !== 200) {
+                    req.log.error({ ...logEntry, requestBody: req.body }, "Error response.");
+                }
+                // right now we log the request body as well even on a success response
+                // this probably isnt sutainable in the long term, but it should help us
```
bigspider: typos `isnt`, `sutainable`

src/dataEntities/appointment.ts
yahgwai: should have been used - added now

src/dataEntities/appointment.ts
yahgwai: Yeah, updated.

src/service.ts
yahgwai: Updated

src/dataEntities/appointment.ts
bigspider: The `parse` function still doesn't have the default parameter. Moreover, once there is the default parameter, you can replace `(log || logger)` with just `log`.

PR #290, opened 2019-08-27
Miscellaneous changes from testing

yahgwai:
Many miscellaneous updates based on feedback from the testing - unfortunately the changes are well documented/separated which makes this PR hard to review. Do your best!

I also made some changes in the block processor which are probably worth looking at quite closely.


src/contractInfo/dataRegistry.ts
```diff
@@ -0,0 +1,366 @@
...
...MPDEST SWAP1 POP SWAP2 SWAP1 POP JUMP JUMPDEST PUSH1 0x0 PUSH3 0x19F4 DUP3 PUSH3 0x1960 JUMP JUMPDEST SWAP1 POP SWAP2 SWAP1 POP JUMP JUMPDEST DUP3 DUP2 DUP4 CALLDATACOPY PUSH1 0x0 DUP4 DUP4 ADD MSTORE POP POP POP JUMP JUMPDEST PUSH1 0x0 JUMPDEST DUP4 DUP2 LT ISZERO PUSH3 0x1A2A JUMPI DUP1 DUP3 ADD MLOAD DUP2 DUP5 ADD MSTORE PUSH1 0x20 DUP2 ADD SWAP1 POP PUSH3 0x1A0D JUMP JUMPDEST DUP4 DUP2 GT ISZERO PUSH3 0x1A3A JUMPI PUSH1 0x0 DUP5 DUP5 ADD MSTORE JUMPDEST POP POP POP POP JUMP JUMPDEST PUSH1 0x0 PUSH1 0x1F NOT PUSH1 0x1F DUP4 ADD AND SWAP1 POP SWAP2 SWAP1 POP JUMP INVALID PUSH1 0x80 PUSH1 0x40 MSTORE CALLVALUE DUP1 ISZERO PUSH2 0x10 JUMPI PUSH1 0x0 DUP1 REVERT JUMPDEST POP PUSH1 0x40 MLOAD PUSH1 0x20 DUP1 PUSH2 0x109F DUP4 CODECOPY DUP2 ADD DUP1 PUSH1 0x40 MSTORE PUSH2 0x2F SWAP2 SWAP1 DUP2 ADD SWAP1 PUSH2 0x91 JUMP JUMPDEST DUP1 PUSH1 0x0 DUP2 SWAP1 SSTORE POP CALLER PUSH1 0x1 PUSH1 0x0 PUSH2 0x100 EXP DUP2 SLOAD DUP2 PUSH20 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF MUL NOT AND SWAP1 DUP4 PUSH20 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF AND MUL OR SWAP1 SSTORE POP POP PUSH2 0xC4 JUMP JUMPDEST PUSH1 0x0 PUSH2 0x89 DUP3 MLOAD PUSH2 0xBA JUMP JUMPDEST SWAP1 POP SWAP3 SWAP2 POP POP
```
bigspider: Perhaps this could be a separate ticket, but I guess we could add the generation of this kind of files to the build flow, instead of committing the bytecode?

src/dataEntities/appointment.ts
```diff
@@ -271,9 +283,21 @@ export class Appointment {
...
+        pisaContractAddress: string,
+        log: Logger = logger
+    ) {
         if (this.paymentHash.toLowerCase() !== Appointment.FreeHash) throw new PublicDataValidationError("Invalid payment hash."); // prettier-ignore
 
+        // the start block must be close to the current block +/- 5
+        const currentHead = blockCache.head.number;
+        if(this.startBlock < (currentHead - 5)) throw new PublicDataValidationError(`Start block too low. Start block must be within 5 blocks of the current block ${currentHead}.`); // prettier-ignore
+        if(this.startBlock > (currentHead + 5)) throw new PublicDataValidationError(`Start block too high. Start block must be within 5 blocks of the current block ${currentHead}.`); // prettier-ignore
+        if((this.endBlock - this.startBlock) > 60000) throw new PublicDataValidationError(`Appointment duration too great. Maximum duration between start and block is 60000.`); // prettier-ignore
```
bigspider: typo, "between start and block" => "between start and end block"

src/dataEntities/appointment.ts
```diff
@@ -271,9 +283,21 @@ export class Appointment {
...
+        if((this.endBlock - this.startBlock) < 100) throw new PublicDataValidationError(`Appointment duration too small. Minimum duration between start and block is 100.`); // prettier-ignore
+        
+
         try {
```
bigspider: Same typo as above.

src/dataEntities/appointment.ts
```diff
@@ -271,9 +283,21 @@ export class Appointment {
...
+        log: Logger = logger
+    ) {
         if (this.paymentHash.toLowerCase() !== Appointment.FreeHash) throw new PublicDataValidationError("Invalid payment hash."); // prettier-ignore
 
+        // the start block must be close to the current block +/- 5
+        const currentHead = blockCache.head.number;
+        if(this.startBlock < (currentHead - 5)) throw new PublicDataValidationError(`Start block too low. Start block must be within 5 blocks of the current block ${currentHead}.`); // prettier-ignore
+        if(this.startBlock > (currentHead + 5)) throw new PublicDataValidationError(`Start block too high. Start block must be within 5 blocks of the current block ${currentHead}.`); // prettier-ignore
+        if((this.endBlock - this.startBlock) > 60000) throw new PublicDataValidationError(`Appointment duration too great. Maximum duration between start and block is 60000.`); // prettier-ignore
+        if((this.endBlock - this.startBlock) < 100) throw new PublicDataValidationError(`Appointment duration too small. Minimum duration between start and block is 100.`); // prettier-ignore
```
bigspider: Same typo as above

src/dataEntities/appointment.ts
```diff
@@ -271,9 +283,21 @@ export class Appointment {
...
+        
+
```
bigspider: The point of hiring Pisa is that you can go offline immediately after; but if you use the natural approach of using the current block as `startBlock` and then going offline, an attacker capable of a re-org of a few blocks might cause the trigger to happen before the `startBlock`, and Pisa wouldn't be watching.

Perhaps the `startBlock` should be allowed to be more in the past; for example if we consider _d_  to be the reorg-safe depth, then we could allow up to `currentBlock - d - 5` in the past.

Any specific reason we limit the `startBlock` to not be far in the future? I might have hired another watchtower for the next 1000 blocks, so I want to hire Pisa from block `currentBlock + 1000`. I don't see issues with that (although we probably don't want appointments that _really_ far in the future).

src/responder/component.ts
```diff
@@ -54,8 +54,13 @@ export class ResponderAppointmentReducer implements StateReducer<ResponderAppoin
             if (!tx.to) continue;
 
             // look for matching transactions
-            const txIdentifier = new PisaTransactionIdentifier(tx.chainId, tx.data, tx.to, tx.value, tx.gasLimit);
+            // we should use chain id, but for some reason chain id is not present in transactions from ethersjs
+            // therefore we fallback to network id when chain id is not present
+            const txChainId = tx.chainId || (tx as any).networkId;
```
bigspider: It looks like ethers.js declares `chainId` in the type, but it does not sanitizes if it receives `networkId` from the nodes (not sure why the inconsistency).
Not a big fan of having this here, I would rather sanitize it in the BlockFactory of the BlockProcessor, setting the chainId as we expect if networkId is set instead.

src/responder/multiResponder.ts
```diff
@@ -63,16 +64,16 @@ export class MultiResponder {
...
                 const txIdentifier = new PisaTransactionIdentifier(
                     this.chainId,
-                    appointment.data,
-                    appointment.contractAddress,
+                    appointment.encodeForResponse(),
+                    this.pisaContractAddress,
                     new BigNumber(0),
-                    appointment.gasLimit
+                    // it appears that sometimes pisa requires a lot of gas to function - resetting data shards?
+                    appointment.gasLimit.add(400000)
```
bigspider: Seems a bit hackish, perhaps we should have an open issue to keep track.

src/service.ts
```diff
@@ -58,7 +63,7 @@ export class PisaService extends StartStopService {
         this.applyMiddlewares(app, config);
 
         // block cache and processor
-        const cacheLimit = config.maximumReorgLimit === undefined ? 200 : config.maximumReorgLimit;
+        const cacheLimit = !config.maximumReorgLimit ? 200 : config.maximumReorgLimit;
```
bigspider: In general, for "primitive or null" and "primitive or undefined" we should prefer comparing with "== null" or "== undefined", as the comparison would fail if config.maximumReorgLimit is 0 (though not a valid choice in this case).

test/smoke/alpha.test.ts
```diff
@@ -0,0 +1,124 @@
...
+        });
+        console.log(response)
+        
+        const tx = await rescueContract.help()
+        console.log("help broadcast")
+        await tx.wait();
+        console.log("help mined")
+        return;
+    }).timeout(100000);
+});
```
bigspider: Is this test incomplete? Doesn't seem to be waiting for Pisa's response.

test/src/dataEntities/appointment.test.ts
```diff
@@ -189,7 +204,19 @@ describe("Appointment", () => {
         const testAppointment = Appointment.parse(clone);
         const signedAppointment = await sign(testAppointment, customerSigner);
 
-        await signedAppointment.validate();
+        await signedAppointment.validate(blockCache, pisaContractAddress);
+    });
+
+    // TODO:274:should not be able to specufy "uint8[]", [[2]] - with a 2 in here
+    fnIt<Appointment>(a => a.validate, "can specify none of the indexed arguments", async () => {
```
bigspider: TODO

src/contractInfo/dataRegistry.ts
yahgwai: Yeah, I'll make a ticket

src/dataEntities/appointment.ts
yahgwai: Done

src/dataEntities/appointment.ts
yahgwai: Done

src/dataEntities/appointment.ts
yahgwai: yeah, I figured 5 was enough depth for that.

In the equation currentBlock - d - 5, what is the purpose of the 5? Just margin of error, different clocks? For currentBlock - d - m what values would you choose? I suppose d would have to be around 20? and m 1 or 2?

For not fair in the future, the reason is just because we're still early stage, and free. I dont want us to bank up a load of unpaid appointments forever - so I picked some limit.

src/responder/component.ts
yahgwai: Hmm, yeah, the chainid thing is really annoying tbh.

The real problem I have here is that network id and chain id are actually different - I've probably made a bad decision to use network id as a fallback. Perhaps a better way would be to get the chain id from the provider, since we know if we got a transaction from there it must have the same chain id.

src/responder/multiResponder.ts
yahgwai: Yeah, I agree. #292

src/service.ts
yahgwai: Yeah, I agree, 0 isnt a valid choice - but I'll check that somewhere else instead.

src/dataEntities/appointment.ts
bigspider: I think 5 is enough against accidental forks, but definitely not against active attackers.

for "startBlock >= currentBlock - d - m" my idea is the the customer sets `startBlock = currentBlock - d` so `m` should be chosen so that the chance of `m` block being mined during the negotiation of the appointment is negligible; so 5 might be good. Although another concern is that the client might be using a node which fell behind (probably quite common) - which could suggest a significantly larger value of `m` in practice, or some other form of mitigation (but this is definitely out of scope for this ticket).

`d` should be the "no fork bigger than this will ever happen" threshold, so perhaps 20 or 30; we actually have the `maximum-reorg-limit` setting for Pisa, although it might make sense to have a smaller limit in the appointment validation (e.g. we can internally handle bigger reorgs than we declare).

We also want `d + m` to be much smaller than the `maxDepth` of the blockCache (and perhaps we should enforce this explicitly).

So my suggestion is `d = 20` and `m = 5` as good values for now.

src/responder/component.ts
bigspider: I've seen ethers.js' code during the review, and there was a place were networkId was assigned from chainId, or vice versa - so yeah, it might be better to avoid relying on ethers.js for this!

test/smoke/alpha.test.ts
yahgwai: Yep this was missing the check! This is just a smoke test to run against ropsten - updated now, but still wont work at the moment as I still need to deploy the pisa contract on ropsten.

test/src/dataEntities/appointment.test.ts
yahgwai: It turns out there's no easy way to do this - the only way I can think of is to parse the arguments one by one, substringing the rest as we go. I'll make a new ticket for this, but I'm actually considering changing this encoding method, perhaps it would be better to just encode the topic client side. This would give us less control over what events we could watch, but I dont think that would matter. #293

src/responder/component.ts
yahgwai: I've kept the code and moved it down into the block factory. Although it is dodgy to interchange the two, as long as we never point at Ethereum Classic or any other forks this shouldnt be a problem. Those sound like famous last words lol.

src/dataEntities/appointment.ts
yahgwai: Ive done d=20 m=3

src/watcher/store.ts
```diff
@@ -3,19 +3,23 @@ import { LevelUp } from "levelup";
 import encodingDown from "encoding-down";
 import { LockManager } from "../utils/lock";
 import { Appointment } from "../dataEntities/appointment";
+import EncodingDown from "encoding-down";
```
bigspider: Missed this before: double import for encoding-down.

yahgwai: Lol, ty

Issue #289, opened 2019-08-27
Add persistence to the BlockchainMachine


bigspider:
When Pisa crashes or is restarted, the Components should always be able to recover a valid past anchor state and compute the most recent state accordingly, so that no event can be missed.

The recovery process should work even if the last checkpointed state is associated to blocks that are deeper than the maxDepth of the BlockCache.

bigspider:
Related issue: #179, might become redundant after solving this.

bigspider:
Saving only the last anchor state would make Pisa unable to correctly handle a re-org that happened during the crash. On the other hand, saving a state that is too old might lead to Pisa unnecessarely repeating actions, which would at least cause a waste of gas.

The most sound approach might be to save the state for each block; in order to avoid a blowup in storage, one could save the state at some old block (deep enough that forks are not a concern), and then store incremental 'diffs' for each subsequent block.

A simple approach for now could be to store all the states for the latest blocks, but then after (say) 4 blocks prune exponentially more states, so that the number of stored anchor states is bounded by `log(maxDepth)` + a small constant.

Issue #179, opened 2019-06-03
BlockCache should repopulate from disk on recovery


bigspider:
Currently, the BlockCache always starts empty and never stores any block deeper than the first block. In case of crash recovery, it is necessary to fill to maxDepth anyway; moreover, blocks should be stored on disk in order to make recovery fast (as the provider might be remotely accessed).

PR #315, opened 2019-09-06
Add persistence to the BlockchainMachine

bigspider:
New stores added:
- BlockItemStore: generic store used by both the BlockchainMachine and the BlockCache to store blocks and informations on them, while allowing efficient pruning.
- BlockProcessorStore: to keep track of the current head of the blockchain in the BlockProcessor
- ActionStore: used to persist the per-component actions that are generated for each component to be able to restart them in case of crash.

Incidentally, moved the new_block event into the BlockCache instead of the BlockProcessor, that instead still retains the new_head event. The new_block event is now never executed multiple times for the same block (even in case of reorg), as each block is added to the BlockCache only once.

Opening separate issues for:
- The BlockItemStore is used in a non-typesafe way by the other classes, and it's too tightly coupled with them (especially with the BlockCache); moreover, several classes using it need to be sure not to overwrite each other's keys. It will need a refactoring to be more robust. Moreover, BlockItemStore should allow to batch updates to execute them atomically (like the underlying db).
- The events in the BlockProcessor and in the BlockCache are now handled with ad-hoc code instead of the default EventEmitter. This has the advantage of allowing asynchronous handling and type-safe listeners, but there is duplication of some event-handling logic (addListener/removeListener) that might be useful to extract into a separate utility class.

Other incidental changes (not related to the original issue):
- changed the maximum line length for prettier from 120 to 160.
- made the last param of `fnIt` optional, like for `it` (to have suspended tests).

Closes: #289 
Closes: #179


src/blockMonitor/blockchainMachine.ts
```diff
@@ -31,62 +106,92 @@ export class BlockchainMachine<TBlock extends IBlockStub> extends StartStopServi
...
-            } else {
-                newState = component.reducer.getInitialState(block);
-            }
+                let newState: AnchorState;
+                let prevHeadAnchorState: AnchorState | null = null;
+                if (this.blockProcessor.blockCache.hasBlock(block.parentHash)) {
+                    const parentBlock = this.blockProcessor.blockCache.getBlock(block.parentHash);
 
-            states.set(block, newState);
+                    if (parentBlock) {
```
yahgwai: No need to check if parentBlock exists here.

src/blockMonitor/blockchainMachine.ts
```diff
@@ -31,62 +106,92 @@ export class BlockchainMachine<TBlock extends IBlockStub> extends StartStopServi
...
+                        prevHeadAnchorState = this.blockItemStore.getItem(parentBlock.hash, `${component.name}:prevEmittedState`);
```
yahgwai: I think you need to include a good explanation somewhere as to why we need to store the prevEmittedState for each component against every block.

src/blockMonitor/blockchainMachine.ts
```diff
@@ -31,62 +106,92 @@ export class BlockchainMachine<TBlock extends IBlockStub> extends StartStopServi
...
+                    }
+
+                    const prevAnchorState =
+                        this.blockItemStore.getItem(parentBlock.hash, `${component.name}:state`) || component.reducer.getInitialState(parentBlock);
```
yahgwai: All item keys like `state` and `attached` etc. should be public static readonly strings on the BlockItemStore

src/dataEntities/block.ts
```diff
@@ -50,3 +54,87 @@ export interface ResponderBlock extends IBlockStub {
...
+
+            const itemsAtHeight = this.itemsByHeight.get(height);
+            if (itemsAtHeight) itemsAtHeight.add(memKey);
+            else this.itemsByHeight.set(height, new Set([memKey]));
+            this.items.set(memKey, value);
+        }
+    }
+    protected async stopInternal() {}
+
+    public async putBlockItem(blockHeight: number, blockHash: string, itemKey: string, item: any) {
```
yahgwai: I know you have another ticket that talks about type safety. Although this isnt a perfect solution to type-safety, I think a very quick and easy fix would be to just add some utility functions like:
```
    private static readonly KEY_ANCHOR_STATE = "ANCHOR_STATE";
    public async putAnchorState(componentName: string, blockNumber: number, blockHash: string, anchorState: AnchorState) {
        await this.putBlockItem(blockNumber, blockHash, `${componentName}:${BlockItemStore.KEY_ANCHOR_STATE}`, anchorState);
    }
    public getAnchorState<TAnchorState>(blockHash: string, componentName: string): TAnchorState {
        return this.getItem(blockHash, `${componentName}:${BlockItemStore.KEY_ANCHOR_STATE}`)
    }
```
You could even make the `putBlockItem` and `getItem` functions private if you like - unless you want to keep them public for testing. I think this would be worth doing on this branch.

Although maybe a better call signature would be:
```
    const anchorState = store.get.anchorState(hash, component.name)
```
since we'll have a lot of getter and putter functions that you could collect on a get/put object inside the store.

src/dataEntities/block.ts
```diff
@@ -50,3 +54,87 @@ export interface ResponderBlock extends IBlockStub {
...
+    /**
+     * Gets the item with key `itemKey` for block `blockHash`.
+     * Returns `undefined` if a key is not present.
+     **/
+    public getItem(blockHash: string, itemKey: string) {
+        const key = `${blockHash}:${itemKey}`;
+        return this.items.get(key);
+    }
+
+    // This behavior is too specific for the needs of the BlockCache; should be refactored
```
yahgwai: This comment seems out of date, or like it belongs on a ticket.

src/dataEntities/block.ts
```diff
@@ -50,3 +54,87 @@ export interface ResponderBlock extends IBlockStub {
...
+
+        const blocks: (TBlock & { attached: boolean })[] = [];
+
+        for (const item of itemsAtHeight) {
+            // check if it is the actual block
+            if (item.endsWith(":block")) {
+                const itemKey = item.slice(0, -":block".length);
+                const block = this.items.get(item) as TBlock;
+                const attached = this.items.get(itemKey + ":attached") as boolean;
+                blocks.push({ ...block, attached: attached });
```
yahgwai: As we discussed before, it's probably safer to return an object with a nested block property to avoid collisions

src/blockMonitor/blockchainMachine.ts
```diff
@@ -31,62 +106,92 @@ export class BlockchainMachine<TBlock extends IBlockStub> extends StartStopServi
...
+
+                // this is now the latest anchor stated for an emitted head block; update the store accordingly
+                await this.blockItemStore.putBlockItem(head.number, head.hash, `${component.name}:prevEmittedState`, state);
+
+                if (prevEmittedState) {
+                    // save actions in the store
+                    const newActions = component.detectChanges(prevEmittedState, state);
+                    if (newActions.length > 0) await this.actionStore.storeActions(component.name, newActions);
+
+                    // load all the actions (might include oldler actions; also, they now have id)
```
yahgwai: If we're loading older actions isnt it possible that if we have to quick consecutive blocks we fire all the actions for the component twice?

src/blockMonitor/blockchainMachine.ts
```diff
@@ -31,62 +106,92 @@ export class BlockchainMachine<TBlock extends IBlockStub> extends StartStopServi
...
+                // If the parent is available and its anchor state is known, the state can be computed with the reducer.
+                // If the parent is available but its anchor state is not known, first compute its parent's initial state, then apply the reducer.
+                // Finally, if the parent is not available at all in the block cache, compute the initial state based on the current block.
 
-                newState = component.reducer.reduce(prevAnchorState, block);
-            } else {
-                newState = component.reducer.getInitialState(block);
-            }
+                let newState: AnchorState;
+                let prevHeadAnchorState: AnchorState | null = null;
```
yahgwai: Rather than having a nullable prevAnchorState, I'd be tempted to set prevAnchorState = newAnchorState where the parent block isnt available. Since the only case where this could happen is when PISA starts for the first time, and in that case the only consequence could only be to pass the newAnchorStae and prevAnchorState to detect changes, which should have negligible effect. If you cover/document this one edge case then you never need to worry about prevAnchorState being null.

test/src/blockMonitor/blockProcessor.test.ts
```diff
@@ -45,42 +53,60 @@ const blocksByHash: { [key: string]: IBlockStub } = {
     c10: { number: 10, hash: "c10", parentHash: "c9" }
 };
 
+describe("BlockProcessorStore", () => {
+    let db: any;
+    let store: BlockProcessorStore;
+
+    beforeEach(async () => {
```
yahgwai: Maybe one more test - "setLastestHeadNumber overwites current head number"

test/src/blockMonitor/blockchainMachine.test.ts
```diff
@@ -50,34 +55,148 @@ type TestAction = {
...
+        },
+        {
+            name: "action2"
+        }
+    ];
+
+    beforeEach(async () => {
+        db = LevelUp(EncodingDown<string, any>(MemDown(), { valueEncoding: "json" }));
+        actionStore = new ActionStore(db);
+        await actionStore.start();
```
yahgwai: I guess it doesnt really matter, cos our current implementation doesnt do anything. But it might be best to call stop() in afterEach()

.prettierrc.js
```diff
@@ -1,7 +1,7 @@
...
-    semi:  true,
-    trailingComma:  'none',
-    singleQuote:  false,
-    printWidth:  120,
-    tabWidth:  4
+module.exports = {
+    semi: true,
+    trailingComma: 'none',
+    singleQuote: false,
+    printWidth: 160,
```
yahgwai: I think in future we should do formatting changes in a separate PR. It made it quite tough for me to see some of the actual changes on this PR, especially in the tests

src/blockMonitor/blockCache.ts
```diff
@@ -212,7 +222,29 @@ export class BlockCache<TBlock extends IBlockStub> implements ReadOnlyBlockCache
...
+    public addNewBlockListener(listener: NewBlockListener<TBlock>) {
+        this.newBlockListeners.push(listener);
+    }
+
+    /**
+     * Removes `listener` from the list of listeners for new block events.
+     */
+    public removeNewBlockListener(listener: NewBlockListener<TBlock>) {
+        const idx = this.newBlockListeners.findIndex(l => l === listener);
+        if (idx === -1) throw new ApplicationError("No such listener exists.");
```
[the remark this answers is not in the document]
bigspider: I didn't because `listener` is a function; I guess it doesn't harm to print it, though.

src/blockMonitor/blockCache.ts
```diff
@@ -111,90 +107,104 @@ export class BlockCache<TBlock extends IBlockStub> implements ReadOnlyBlockCache
...
+    private async processDetached(height: number) {
+        const blocksAtHeight = this.blockStore.getBlocksAtHeight(height);
+        const blocksToAdd = blocksAtHeight.filter(b => !b.attached);
 
-            // Remove block from detachedBlocksByHash, add to blocksByHash
-            this.blocksByHash.set(blockHash, block);
-            this.detachedBlocksByHash.delete(blockHash);
+        for (const block of blocksToAdd) {
+            await this.blockStore.putBlockItem(height, block.hash, "attached", true);
+            block.attached = true;
```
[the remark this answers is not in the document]
bigspider: Indeed, `block.attached = true;` is redundant. Removed.

src/blockMonitor/blockCache.ts
```diff
@@ -111,90 +107,104 @@ export class BlockCache<TBlock extends IBlockStub> implements ReadOnlyBlockCache
...
+    private async emitNewBlockEvent(block: TBlock) {
+        for (const listener of this.newBlockListeners) {
+            await listener(block);
+        }
+    }
 
-        for (const blockHash of blockHashesToAdd) {
-            const block = this.detachedBlocksByHash.get(blockHash)!;
+    // Processes all the detached blocks at the given height, marking them as attached if necessary.
+    // If there are any, add them and repeat with the next height (as some blocks might now have become attached)
```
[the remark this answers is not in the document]
bigspider: I assumed that the /** */ comments was for documentation of everything public and it could be picked up by some automated documentation generator to produce a developer manual. No problem switching to /** */ by default for all methods from now on, though.

src/blockMonitor/blockCache.ts
```diff
@@ -111,90 +107,104 @@ export class BlockCache<TBlock extends IBlockStub> implements ReadOnlyBlockCache
...
 
-        // Index block by its height
-        const hashesByHeight = this.blockHashesByHeight.get(block.number);
-        if (hashesByHeight == undefined) {
-            this.blockHashesByHeight.set(block.number, new Set([block.hash])); // create new Set
-        } else {
-            hashesByHeight.add(block.hash); // add to existing Set
-        }
+            if (this.canAttachBlock(block)) {
+                // TODO: this should happen atomically
```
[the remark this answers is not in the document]
bigspider: Removed, still left a comment to remember it is to be done as part of #316.

src/blockMonitor/blockCache.ts
```diff
@@ -111,90 +107,104 @@ export class BlockCache<TBlock extends IBlockStub> implements ReadOnlyBlockCache
...
-            return BlockAddResult.Added;
-        } else {
-            this.detachedBlocksByHash.set(block.hash, block);
-            return BlockAddResult.AddedDetached;
+                // If the minHeight increased, this could also make some detached blocks ready to be attached
+                // This makes sure that they are attached if necessary
+                await this.processDetachedBlocksAtMinHeight();
+                return BlockAddResult.Added;
+            } else {
+                // TODO: this should happen atomically
```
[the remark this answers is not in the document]
bigspider: As above.

src/blockMonitor/blockProcessor.ts
```diff
@@ -104,26 +129,11 @@ export class BlockProcessor<TBlock extends IBlockStub> extends StartStopService
...
-    public static readonly NEW_BLOCK_EVENT = "new_block";
-
     constructor(
         private provider: ethers.providers.BaseProvider,
         blockFactory: BlockFactory<TBlock>,
-        blockCache: BlockCache<TBlock>
+        blockCache: BlockCache<TBlock>,
+        private readonly store: BlockProcessorStore
     ) {
         super("block-processor");
```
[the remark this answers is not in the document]
bigspider: processBlockNumber is still the listener for the provider's "block" event, so no change there.

But even for callbacks to our internal events, since they will be called as
   `   listener(...);`
somewhere, the `this` inside the listener would be wrong without that (unless they are declared with arrow notation, of course).
Essentially the bind should be done for any callback (except perhaps small pure functions that don't use `this`, but they would probably be arrow functions anyway).

src/blockMonitor/blockProcessor.ts
```diff
@@ -79,11 +81,32 @@ export const blockFactory = (provider: ethers.providers.Provider) => async (
...
+    constructor(db: LevelUp<EncodingDown<string, any>>) {
+        this.subDb = sub(db, `block-processor`, { valueEncoding: "json" });
+    }
+
+    async getLatestHeadNumber() {
+        try {
+            const headObj = await this.subDb.get("head");
+            return (headObj as { head: number }).head;
+        } catch (doh) {
+            // TODO: errors other than "key not found" should be logged
```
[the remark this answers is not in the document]
bigspider: Addressed.

src/blockMonitor/blockchainMachine.ts
```diff
@@ -1,26 +1,101 @@
...
-        this.blockProcessor.off(BlockProcessor.NEW_HEAD_EVENT, this.processNewHead);
-        this.blockProcessor.off(BlockProcessor.NEW_BLOCK_EVENT, this.processNewBlock);
+        this.blockProcessor.removeNewHeadListener(this.processNewHead);
+        this.blockProcessor.blockCache.removeNewBlockListener(this.processNewBlock);
     }
 
-    constructor(private blockProcessor: BlockProcessor<TBlock>) {
+    constructor(private blockProcessor: BlockProcessor<TBlock>, private actionStore: ActionStore, private blockItemStore: BlockItemStore<TBlock>) {
         super("blockchain-machine");
         this.processNewHead = this.processNewHead.bind(this);
```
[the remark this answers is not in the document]
bigspider: Yes, as per my comment above.

src/blockMonitor/blockProcessor.ts
```diff
@@ -185,17 +199,32 @@ export class BlockProcessor<TBlock extends IBlockStub> extends StartStopService
...
+
     // Processes a new block, adding it to the cache and emitting the appropriate events
     // It is called for each new block received, but also at startup (during startInternal).
     private async processBlockNumber(blockNumber: number) {
         try {
-            const observedBlock = await this.getBlockRemote(blockNumber);
+            // we cant process blocks greater than max depth of the cache
+            // so if the cache is empty any block is fine, otherwise
+            // the blocknumber cannot be more than maxDepth greater than the head
+            const maxBlock = this.mBlockCache.isEmpty // prettier-ignore
```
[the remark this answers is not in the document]
bigspider: While it does not matter logically, it is still better to split the remaining blocks to be added in chunks, so that the number of blocks kept unattahed in the BlockCache is bounded. Otherwise _all_ the blocks from the stored snapshot to the current head are kept in memory until they can be attached, since they can't be pruned before that.

src/blockMonitor/blockProcessor.ts
```diff
@@ -135,45 +145,49 @@ export class BlockProcessor<TBlock extends IBlockStub> extends StartStopService
...
+    private async processNewHead(headBlock: Readonly<TBlock>) {
+        try {
+            this.mBlockCache.setHead(headBlock.hash);
 
-            // Emit all the blocks past the latest block in the ancestry that was emitted as head
-            // In case of re-orgs, some blocks might be re-emitted multiple times.
-            for (const block of blocksToEmit) {
-                this.emit(BlockProcessor.NEW_BLOCK_EVENT, block);
-            }
+            await this.store.setLatestHeadNumber(headBlock.number);
```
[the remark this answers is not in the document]
bigspider: Done.

Do you think we should add any handling for potenial errors thrown by the listeners?
Currently, the db write happens in the try block, perhaps it shoud be outside (or better in a separate try/catch block), so that errors in listeners are logged, but don't mess upd with the updates.

src/blockMonitor/blockProcessor.ts
```diff
@@ -135,45 +145,49 @@ export class BlockProcessor<TBlock extends IBlockStub> extends StartStopService
...
+
+            // only emit new head events after it is started
+            if (this.started) {
+                // Emit the new head
+                await Promise.all(this.newHeadListeners.map(listener => listener(headBlock)));
 
-            // Emit the new head
-            this.emit(BlockProcessor.NEW_HEAD_EVENT, headBlock, nearestEmittedHeadInAncestry);
-            this.emittedBlockHeads.add(headBlock);
+                this.emittedBlockHeads.add(headBlock);
```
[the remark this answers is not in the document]
bigspider: No, removed!

src/blockMonitor/blockchainMachine.ts
```diff
@@ -1,26 +1,101 @@
...
+
+            const componentActions = this.actions.get(componentName);
+            if (componentActions) componentActions.add(actionWithId);
+            else this.actions.set(componentName, new Set([actionWithId]));
+        }
+    }
+    protected async stopInternal() {}
+
+    // commodity function to wrap an action with an existing id, or generate a new unique id if not given
+    private addId(action: ComponentAction, id?: string): ActionAndId {
```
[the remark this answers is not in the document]
bigspider: Done.

src/blockMonitor/blockchainMachine.ts
```diff
@@ -31,62 +106,92 @@ export class BlockchainMachine<TBlock extends IBlockStub> extends StartStopServi
...
+
+                    const prevAnchorState =
+                        this.blockItemStore.getItem(parentBlock.hash, `${component.name}:state`) || component.reducer.getInitialState(parentBlock);
+
+                    newState = component.reducer.reduce(prevAnchorState, block);
+                } else {
+                    newState = component.reducer.getInitialState(block);
+                }
+
+                // states.set(block, newState);
```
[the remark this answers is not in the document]
bigspider: Oops, done.

src/blockMonitor/blockchainMachine.ts
bigspider: Done.

src/blockMonitor/blockchainMachine.ts
bigspider: Done.

I am not sure I like having these keys in the BlockItemStore (meaning that the BlockItemStore has to know about all the keys used by other classes), but that could be part of the improvements in #316.

src/blockMonitor/blockchainMachine.ts
bigspider: Added in 2b390d7 as documentation of the BlockchainMachine class.

src/dataEntities/block.ts
bigspider: Removed, it might likely be part of #316 anyway.

src/dataEntities/block.ts
bigspider: Done in 41b88b5. Incidentally, it also makes other code more readable, as removing the "attached" field was a pain before and now it's not needed.

src/blockMonitor/blockchainMachine.ts
bigspider: As processNewBlock and processNewHead in the BlockchainMachine now acquire a lock, it is guaranteed that events processing never happen in parallel in the BlockchainMachine, so we should be good - unless I missed something.

src/blockMonitor/blockCache.ts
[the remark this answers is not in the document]
yahgwai: Ah, fair enough i misread. I dont think you should print the function - it'll just print [ object Object ] or something silly.

src/dataEntities/block.ts
bigspider: Done in 1294449, with syntax "store.{block|attached|anchorState|etc.}.get/set".

src/blockMonitor/blockchainMachine.ts
bigspider: Keys now all in BlockItemStore and private.

src/blockMonitor/blockchainMachine.ts
bigspider: There are only two places where this matters, that is the new-block and the new-head event handlers, and there shouldn't be more in the future; hence leaving the nullability for now.

test/src/blockMonitor/blockProcessor.test.ts
bigspider: Done in c24a745.

test/src/blockMonitor/blockchainMachine.test.ts
bigspider: Done in 6845412, with a guard to prevent stopping the servicecs twice, as some tests already stop some services prematurely.

.prettierrc.js
bigspider: Sorry about that; I disabled the formatOnSave option for prettier.

src/blockMonitor/blockCache.ts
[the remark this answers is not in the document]
bigspider: From some tests on the console it might show the source code; not sure though, and perhaps it would be silly anyway. Reverted in cb7a8ab.

src/blockMonitor/blockchainMachine.ts
```diff
@@ -1,26 +1,101 @@
...
+import EncodingDown from "encoding-down";
+const sub = require("subleveldown");
+import uuid = require("uuid/v4");
+
+export interface ActionAndId {
+    id: string;
+    action: ComponentAction;
+}
+
+export class ActionStore extends StartStopService {
```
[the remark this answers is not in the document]
bigspider: Done in 36d11cf.

bigspider:
Added some initial documentation to the BlockItemStore in 97d52ab.

src/blockMonitor/blockchainMachine.ts
bigspider: I see, what I was missing is that indeed the actions run in their separate "thread". Good point.
As the issue is quite serious (and perhaps easy to miss), I spent some time to create a failing testcase in 75a3572cbeb2e798f373368eaa91caa4a81f478e, which I then fixed in 53c1919432c2fc28d99381d83194aa5cad7bd9f8 by keeping track of the ids of running actions.

src/dataEntities/block.ts
```diff
@@ -50,3 +55,145 @@ export interface ResponderBlock extends IBlockStub {
...
+        const itemsAtHeight = this.itemsByHeight.get(height) || new Set();
+        // collect blocks and attachment info
+
+        const blocks: BlockAndAttached<TBlock>[] = [];
+
+        for (const item of itemsAtHeight) {
+            // check if it is the actual block
+            if (item.endsWith(`:${BlockItemStore.KEY_BLOCK}`)) {
+                const itemKey = item.slice(0, -`:${BlockItemStore.KEY_BLOCK}`.length);
+                const block = this.items.get(item) as TBlock;
```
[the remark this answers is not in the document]
bigspider: Done in ae3ba4d.

src/blockMonitor/blockProcessor.ts
```diff
@@ -81,21 +90,45 @@ export const blockFactory = (provider: ethers.providers.Provider) => async (
...
+        this.logger = createNamedLogger(`block-processor-store`);
+    }
+
+    async getLatestHeadNumber() {
+        try {
+            const headObj = await this.subDb.get("head");
+            return (headObj as { head: number }).head;
+        } catch (doh) {
+            // Rethrow any error, except for "key not found", which is expected
+            if (doh.type === "NotFoundError") return undefined;
```
yahgwai: What is the .type property? Does that exist on NotFoundErrors?

src/blockMonitor/blockchainMachine.ts
bigspider: Further refactoring done in f99b463, including some minor refactoring and documentation of the ActionStore.

src/blockMonitor/blockchainMachine.ts
```diff
@@ -1,26 +1,64 @@
...
+     * Runs all the actions in `actionAndIds` for `component`. Actions are all executed in parallel, and each action is removed
+     * from the ActionStore upon completion.s
+     */
+    private runActionsForComponent(component: Component<AnchorState, IBlockStub, ComponentAction>, actionAndIds: Iterable<ActionAndId>) {
+        // Side effects must be thread safe, so we can execute them concurrently
+        // Note that actions are executed in background and not awaited for in here.
+        [...actionAndIds].forEach(async a => {
+            try {
+                await component.applyAction(a.action);
+                this.actionStore.removeAction(component.name, a);
```
yahgwai: Shouldnt we be awaiting the removeAction?

bigspider: Ooops, funny I added the try/catch exactly to log errors thrown from the removeAction, but I forgot to await it. >:)
Fixed in 2f123df.

PR #324, opened 2019-09-12
Appointment serialised directly as a struct

yahgwai:
- [x] Tests no longer require ganache to be started first - instead the truffle config starts one itself using ganache core
- [x]  PISAHash.sol now accepts a struct argument
- [x]  Updated tests and related code
- [x]  Updated documentation


stonecoldpat:
This looks good to me. Excited that even the unit tests were updated

Should save effort when users want to call via the api, so all they need to do is a tuple and not bother with all the encoding effort.

PR #332, opened 2019-09-13
Client lib

yahgwai:
Added a thin client lib and published it to npm (doesnt work with the current version of the API at alpha.pisa.watch since that's running older code).


client/index.ts
```diff
@@ -0,0 +1,353 @@
...
+                      id,
+                      nonce,
+                      startBlock,
+                      endBlock,
+                      contractAddress,
+                      data,
+                      gasLimit,
+                      challengePeriod
+                  );
+        return await this.executeRequest(request);
```
bigspider: If some but not all three optional params, this will silently ignore the other two without giving an error.
I guess you can just call `generateRequest` with all parameters, instead?

test/integration/remote/endToEnd.test.ts
```diff
@@ -274,38 +225,23 @@ describe("Integration", function() {
...
-            const hashedWithAddress = ethers.utils.keccak256(encoding);
-            const sig = await wallets0[i].signMessage(ethers.utils.arrayify(hashedWithAddress));
-            const clone = { ...appointment, customerSig: sig };
+            const id = "0x0000000000000000000000000000000000000000000000000000000000000001";
 
-            await request.post(`http://localhost:${pisa.config.hostPort}/appointment`, {
-                json: clone
-            });
+            await client.generateAndExecuteRequest(
+                disgest => wallets0[i].signMessage(ethers.utils.arrayify(disgest)),
```
bigspider: typo disgest

test/integration/remote/endToEnd.test.ts
```diff
@@ -157,40 +125,23 @@ describe("Integration", function() {
...
-
-        const clone = { ...appointment, customerSig: sig };
-
-        const res = await request.post(`http://localhost:${pisa.config.hostPort}/appointment`, {
-            json: clone
-        });
+        const id = "0x0000000000000000000000000000000000000000000000000000000000000001";
+
+        await client.generateAndExecuteRequest(
+            disgest => key0.wallet.signMessage(ethers.utils.arrayify(disgest)),
```
bigspider: typo disgest

client/index.ts
yahgwai: There is a check in generateRequest that throws an error. Or do you mean something else?

test/integration/remote/endToEnd.test.ts
yahgwai: Fixed.

client/index.ts
bigspider: You are doing:

```
function foo(params, A?, B?, C?) {
  if (A && B && C) generateRequest(params, A, B, C);
  else generateRequest(params);
}
```
So if for example A is given, but not B and C, then the condition `A && B && C` would be false and `generateRequest(params)` would be called.
I think you should just call `generateRequest(params, A, B, C)`, so that the parameters are actually checked.

yahgwai: I dont mean the check in `generateAndExecuteRequest`, I mean the check **inside** `generateRequest`.

bigspider: I mean that if you call generateAndExecuteRequest with only eventAddress and not eventAbi and eventArgs, then generateAndExecuteRequest would call generateRequest without any of the optional params; so no error would be thrown and the eventAddress parameter would be ignored.
But anyway, since generateRequest has proper checks, you can just do:
```
	        const request = await this.generateRequest(
                      signer,
                      customerAddress,
                      id,
                      nonce,
                      startBlock,
                      endBlock,
                      contractAddress,
                      data,
                      gasLimit,
                      challengePeriod,
                      eventAddress,
                      eventABI,
                      eventArgs
                  );
```
With all arguments.

yahgwai: Ah, I see.

The type system wont actually let me do as you've suggested because it wont accept the type `string | undefined` for any event args (the generateRequest func does only doesnt have that overload available).

So I've had to do some weird casting to strings.

bigspider: I see, indeed that's not ideal. I guess I'd prefer to repeat the argument check independentently in both functions, then, rather than forcing the type system.

If you instead hate the code duplication more, you can further simplify like this:
```
    public async generateAndExecuteRequest(
        signer: (digest: string) => Promise<string>,
        customerAddress: string,
        id: string,
        nonce: number,
        startBlock: number,
        endBlock: number,
        contractAddress: string,
        data: string,
        gasLimit: number,
        challengePeriod: number,
        eventAddress?: string,
        eventABI?: string,
        eventArgs?: string
    ): Promise<AppointmentReceipt> {
        const request = await this.generateRequest(
            signer,
            customerAddress,
            id,
            nonce,
            startBlock,
            endBlock,
            contractAddress,
            data,
            gasLimit,
            challengePeriod,
            // we need to cast to string here to allow the compiler to accept possibly undefined values
            // this should only be the case if a caller has passed in undefined as one of the event args
            // in which case we can pass it to generateRequest in below where an error will be thrown.
            eventAddress as string,
            eventABI as string,
            eventArgs as string
        );
        return await this.executeRequest(request);
    }
```

yahgwai: Fair enought, updated.

PR #333, opened 2019-09-16
Added a custom class for type-safe events

bigspider:
This PR adds a base class Event and a derived concrete class BlockEvent to manage events and event listeners in a type-safe way. This removes some code duplication in the BlockProcessor and the BlockCache.

The base class contains the logic of adding and removing listeners, while concrede implementations implement the `emit` method that allows to call all the listeners for the event. All the listeners are emitted in sequence, and awaited for; thus, when the `emit` method resolves, the caller is guaranteed that all the listeners have been called and returned.

The concrete BlockEvent class implements an event that accepts a single `block` argument.

### DIFF
--- a/src/blockMonitor/blockProcessor.ts
+++ b/src/blockMonitor/blockProcessor.ts
@@ -128,7 +128,7 @@ export class BlockProcessor<TBlock extends IBlockStub> extends StartStopService 
 
     private mBlockCache: BlockCache<TBlock>;
 
-    public NewHead = new BlockEvent<TBlock>();
+    public newHead = new BlockEvent<TBlock>();
 
     // Returned in the constructor by blockProvider: obtains the block remotely (or throws an exception on failure)
     private getBlockRemote: (blockNumberOrHash: string | number) => Promise<TBlock>;
@@ -171,7 +171,7 @@ export class BlockProcessor<TBlock extends IBlockStub> extends StartStopService 
             this.mBlockCache.setHead(headBlock.hash);
 
             // only emit new head events after it is started
-            if (this.started) this.NewHead.emit(headBlock);
+            if (this.started) this.newHead.emit(headBlock);
 
             await this.store.setLatestHeadNumber(headBlock.number);
         } catch (doh) {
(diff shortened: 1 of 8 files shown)


src/blockMonitor/blockProcessor.ts
```diff
@@ -127,7 +128,7 @@ export class BlockProcessor<TBlock extends IBlockStub> extends StartStopService
 
     private mBlockCache: BlockCache<TBlock>;
 
-    private newHeadListeners: NewHeadListener<TBlock>[] = [];
+    public NewHead = new BlockEvent<TBlock>();
```
yahgwai: At the moment I think we're only using capitalised names for static properties

src/utils/event.ts
```diff
@@ -0,0 +1,43 @@
+import { IBlockStub, ApplicationError } from "../dataEntities";
+
+/**
+ * This is a generic class to use as base for concrete classes to manage events and event listeners.
+ * The type parmeter TListener is the type of the event listeners, which should be a function returning a Promise.
+ * Concrete subclasses must implement an `emit` method method to actually emits the event and calls all the listeners,
+ * awaiting each one of them.
+ */
+export abstract class Event<TListener extends Function> {
```
yahgwai: Cant you enforce that Function returns a promise by extending from
```
TListener extends () => Promise<any>
```
or
```
TListener extends () => Promise<TReturn>
```
if you want to consistent return type

bigspider: I didn't know you can constrain versus arrow types like that.
That doesn't work yet, as subclasses (like BlockEvent) might take listeners with a parameter. This seems to work:
```
export abstract class Event<TListener extends (...args: any[]) => Promise<void>> {
```

where I also constrained the return value to `void`, no reason to expect anything else.

src/blockMonitor/blockProcessor.ts
bigspider: I know it's not our convention, but for some reason it felt more natural.
Anyway, changed to lowercase names in 749e18e.

src/utils/event.ts
bigspider: Changed in 749e18e.

PR #345, opened 2019-09-24
Change event ABI + eventArgs to topics

bigspider:


src/public/swagger-doc.json
```diff
@@ -177,7 +171,7 @@
...
...de(\n  [\n    \"uint\", \n    \"uint\", \n    \"uint\", \n    \"uint\", \n    \"uint\", \n    \"uint\", \n    \"bytes32\"\n  ],\n  [\n    request.id,\n    request.nonce,\n    request.startBlock,\n    request.endBlock,\n    request.challengePeriod,\n    request.refund,\n    request.paymentHash\n  ]\n);\n\nconst callBytes = ethers.utils.defaultAbiCoder.encode(\n  [\n    \"address\", \n    \"address\", \n    \"uint\", \n    \"bytes\"\n  ],\n  [\n    request.contractAddress, \n    request.customerAddress, \n    request.gasLimit, \n    request.data\n  ]\n);\n\nconst conditionBytes = ethers.utils.defaultAbiCoder.encode(\n  [\n    \"address\", \n    \"bytes\", \n    \"bytes\", \n    \"bytes\", \n    \"uint\"\n  ],\n  [\n    request.eventAddress,\n    request.topics,\n    request.preCondition,\n    request.postCondition,\n    request.mode\n  ]\n);\n\nconst appointmentBytes = ethers.utils.defaultAbiCoder.encode(\n  [\n    \"bytes\", \n    \"bytes\", \n    \"bytes\"\n  ], \n  [\n    basicBytes, \n    callBytes,\n    conditionBytes\n  ]\n);\n\nconst hashForSig = ethers.utils.keccak256(appointmentBytes);\nconst customerSig = await wallet.signMessage(ethers.utils.arrayify(hashForSig));\n```\n",
```
yahgwai: This documentation doesnt look right. We encode as a tuple now.

bigspider: Actually it wasn't just the documentation, as the encoding in alpha.test.ts wasn't correctly updated (in particular, I was passing the wrong data and format for the new `topics` argument).
Fixed in d969f2f (although I couldn't test it locally, as it runs on the live Pisa). Now checking how to fix the documentation accordingly.
As far as I can tell, this code does not have a counterparty

yahgwai: No, it doesn't.

To be honest, just having to reformat that code occasionally is so painful that I'm considering changing our documentation.

bigspider: Actually, there seems to be a mismatch (even in master) between the encoding used in the client and the encoding used in the alpha test, see https://github.com/PISAresearch/pisa/blob/master/client/index.ts#L50 and https://github.com/PISAresearch/pisa/blob/master/test/smoke/alpha.test.ts#L10.

client/index.ts
```diff
@@ -33,6 +32,16 @@ interface AppointmentReceipt {
     readonly watcherAddress: string;
 }
 
+// Encode the topics in the format expected from Pisa's contract.
```
yahgwai: This should be a private static function on the Client class

bigspider: Done in 17fdde9.

PR #347, opened 2019-09-25
Add get functionality to the API

yahgwai:


src/watcher/store.ts
```diff
@@ -59,23 +69,33 @@ export class AppointmentStore extends StartStopService {
...
-            else await batch.write();
+            await batch.write();
 
-            // add the new appointment
+            // add the new appointment, and replace an old one
             this.mAppointmentsByLocator.set(appointment.locator, appointment);
             this.mAppointmentsById.set(appointment.id, appointment);
+            this.mAppointmentsByCustomerAddress.addToSet(appointment.customerAddress, appointment);
         });
     }
```
bigspider: In previous attempts to write this function, I was splitting it in three blocks:
1) validation (e.g. throw an error if someething is wrong)
2) write the batch to db
3) update the in-memory structures.

The reason is that if writing the batch throws an error, this would make sure that also in-memory changes are not done (while the code now does some in-memory changes before writing the batch).
It's definitely easier to read this version of the code, though (I had to put two conditional blocks over currentAppointment for that reason). Do you think that the in-memory consistency in case of db failure is a desirable property?

yahgwai: I agree, updated.

Issue #352, opened 2019-09-27
The whole service layer lacks good testing


yahgwai:

yahgwai:
Additional to other tests:

- Failure tests for the authentication

PR #353, opened 2019-09-27
Added some authentication to PISA get requests

yahgwai:


src/service.ts
```diff
@@ -337,3 +342,57 @@ export class PisaService extends StartStopService {
...
+            recoveredAddress = ethers.utils.verifyMessage(ethers.utils.arrayify("0x" + blockNumber.toString(16)), signature);
+        } catch (doh) {
+            this.logger.error(doh);
+            throw new PublicDataValidationError("Invalid x-auth-sig header.");
+        }
+        if (recoveredAddress !== customerAddress) throw new PublicDataValidationError("Signing key does not match customer address.");
+    }
+}
+
+class PisaParameters {
```
bigspider: Maybe a more descriptive name like "PisaParametersValidator" or similar.

test/src/serviceEndToEnd.test.ts
```diff
@@ -369,10 +370,11 @@ describe("Service end-to-end", () => {
...
-            KitsuneTools.topics(),
+            KitsuneTools.topics()
         );
 
-        const res = await pisaClient.getAppointmentsByCustomer(account0);
+        const getBlockNumber = await provider.getBlockNumber();
+        const res = await pisaClient.getAppointmentsByCustomer(digest => wallet0.signMessage(arrayify(digest)), account0, getBlockNumber);
         expect(res).to.deep.equal([appointment2.appointment, appointment3.appointment]);
     }).timeout(3000);
 });
```
bigspider: Do you think we should have some tests with failing authentication (missing headers, wrong signature, block more than 5 apart from current head, etc.)?

yahgwai: Yeah, we should. Ideally this would be dont with a unit test though, rather than a end to end. The whole service layer lacks unit tests - I have a ticket her: https://github.com/PISAresearch/pisa/issues/352 and I've added these tests as a bullet point there

yahgwai: I've also added some more end to end tests on this branch:
- block too low
- block too high
- wrong account
- invalid sig

src/service.ts
yahgwai: Renamed

PR #355, opened 2019-09-27
Added an instance name param for additional logging

yahgwai:
This allows us to identify multiple instances on a single logging server


src/dataEntities/config.ts
```diff
@@ -131,6 +131,45 @@ export class ConfigManager {
...
+            throw new Error("Option 'maximum-reorg-limit' cannot be 0.");
+        }
+    }
+}
+
+export class PisaConfigManager {
+    public static getConfig() {
+        const configManager = new ConfigManager(ConfigManager.PisaConfigProperties);
+        return configManager.getConfig()
+    }
```
bigspider: Not sure I get the design of these classes; if ConfigManager is generic while PisaConfigManager is tailored specifically for Pisa, then I think that `PisaConfigProperties` should be moved into `PisaConfigManager`.

yahgwai: Agree, updated

Issue #265, opened 2019-08-12
Fix usages of chai-as-promised in tests


bigspider:
Several tests in the codebase are using chai-as-promised incorrectly, by doing something like:
```
    expect(something).to.eventually.be.rejected;
```
This is incorrect, see https://www.chaijs.com/plugins/chai-as-promised/
It might lead to tests beforee the condition is actually checked.

Those should be either returned (if it is the last check of the test), or awaited explicitly.

yahgwai:
Possible solution:
```
const expectAsync = async <TReturn>(target: Promise<TReturn>) => {
        try {
            return expect(await target);
        } catch (err) {
            return expect(() => {
                throw err;
            });
        }
    };
```
used like this:
```
const face = (await expectAsync(Promise.reject(10))).to.throw;
```

PR #357, opened 2019-10-01
Fixed usages of chaiAsPromised in tests; minor other fix.

bigspider:
Fixed incorrect usages of chai-as-promised.
Incidentally, forced the type in some places where passing the ganache object to things expecting an AsyncSendable, which seems to fail now (they seem to have added some type annotations, making some type checks fail on master).

Closes: #265.


test/endtoend/endToEndSos.test.ts
```diff
@@ -30,7 +30,7 @@ const nextConfig = {
     watcherResponseConfirmations: 0,
     maximumReorgLimit: 10
 };
-const provider = new ethers.providers.Web3Provider(ganache);
+const provider = new ethers.providers.Web3Provider(ganache as ethers.providers.AsyncSendable);
```
yahgwai: You can do the cast once where ganache is declared.

test/endtoend/serviceEndToEnd.test.ts
```diff
@@ -29,7 +30,7 @@ const nextConfig = {
     watcherResponseConfirmations: 0
 };
 
-const provider = new ethers.providers.Web3Provider(ganache);
+const provider = new ethers.providers.Web3Provider(ganache as ethers.providers.AsyncSendable);
```
yahgwai: Same here, probably better to cast at the declaration - I looked into the source code and it seems they;'re missing some types. It looks like the types they're referencing are web3 0.x types whereas the actual included packages are web3 1.x. So maybe they updated the js, but forgot to do the types? Hopefully they'll fix it in the future.

test/endtoend/endToEndSos.test.ts
bigspider: You're right, much cleaner. Done in dde7c9f.

PR #362, opened 2019-10-04
Connext integration

yahgwai:
Added some utility functions to the client


client/index.ts
```diff
@@ -364,5 +371,55 @@ export default class PisaClient {
...
+    /**
+     * Backup some data to the PISA server.
+     * @param signer A function sign the PISA appointment
+     * @param data The data to backup
+     * @param customerAddress The customer address backing up the data
+     * @param startBlock The start block from when the backup should begin - should be within 5 blocks of the current block. Backup will be held for 60,000 blocks
+     * @param id The id for this backup
+     * @param nonce The version of this backup. A backup can be replaced by providing the same backup id but a greater nonce.
+     */
+    public async backUp(signer: (digest: string) => Promise<string>, customerAddress: string, data: string, startBlock: number, id: string, nonce: number) {
```
bigspider: I think I'd prefer backup instead of backUp - not very common.
Also typo in documentation: `signer A function sign the  PISA appointment`, missing "to".

client/package.json
```diff
@@ -1,9 +1,9 @@
 {
   "name": "pisa-client",
-  "version": "0.1.2-rc1",
+  "version": "0.1.4-connext-beta.1",
```
bigspider: I don't think "connext" should be in the version number in master; forgot to remove?

src/blockMonitor/blockchainMachine.ts
```diff
@@ -94,7 +94,7 @@ export class BlockchainMachine<TBlock extends IBlockStub> extends StartStopServi
                     prevHeadAnchorState = this.blockItemStore.prevEmittedAnchorState.get(component.name, parentBlock.hash);
 
                     const prevAnchorState =
-                        this.blockItemStore.anchorState.get<AnchorState>(component.name, parentBlock.hash) ||
+                    this.blockItemStore.anchorState.get<AnchorState>(component.name, parentBlock.hash) ||
```
bigspider: Spacing was right before.

src/service.ts
```diff
@@ -133,7 +135,7 @@ export class PisaService extends StartStopService {
         // set up 404
         app.all("*", function(req, res) {
             res.status(404).json({
-                message: `Route ${req.url} not found, only availale routes are POST at /appointment and GET at /docs.html`
+                message: `Route ${req.url} not found, only available routes are POST at /appointment and GET at /docs.html`
```
bigspider: I was thinking that it could be nicer to serve docs.html at the GET root path. (so that https://alpha.pisa.watch shows the docs as well).

test/src/watcher/watcher.test.ts
```diff
@@ -395,11 +371,11 @@ describe("Watcher", () => {
...
-                items: makeMap(appointment.id, { state: WatcherAppointmentState.WATCHING }),
+                items: { [appointment.id]: { state: WatcherAppointmentState.WATCHING } },
                 blockNumber: 101 + CONFIRMATIONS_BEFORE_REMOVAL
             },
             {
-                items: makeMap(appointment.id, { state: WatcherAppointmentState.WATCHING }),
+                items: { [appointment.id]: { state: WatcherAppointmentState.WATCHING } },
                 blockNumber: 101 + CONFIRMATIONS_BEFORE_REMOVAL + 1
             }
         );
```
bigspider: Can remove makeMap, no longer used.

client/index.ts
yahgwai: Agree, updated

client/package.json
yahgwai: Bumped to 0.1.5

src/blockMonitor/blockchainMachine.ts
yahgwai: Reverted

src/service.ts
yahgwai: Agree, added

test/src/watcher/watcher.test.ts
yahgwai: Removed

PR #376, opened 2019-12-03
Generalized ActionStore to be multipurpose

bigspider:
The actionStore used names that were specific to the Component architecture, but it is just a keyed (sub)-database of sets of objects. This PR makes the type explicit at creation time and makes the names more general.


bigspider:
Renamed in to CachedKeyValueStore in 915db74; also changed all occurrences of "actions" to the generic "items" (still calling the CachedKeyValueStore used by the blockchainMachine "actionStore", though, I think it makes sense there).

packages/block/src/cachedKeyValueStore.ts
```diff
@@ -0,0 +1,84 @@
...
+import uuid = require("uuid/v4");
+
+import { StartStopService } from "@pisa-research/utils";
+
+export interface ItemAndId<TValue> {
+    id: string;
+    value: TValue;
+}
+
+/** This class handles a subdatabase and stores entries that are sets of items indexed by a string key. */
```
yahgwai: These docs should be better.

packages/block/src/blockchainMachine.ts
```diff
@@ -41,15 +41,15 @@ export class BlockchainMachine<TBlock extends IBlockStub> extends StartStopServi
 
     protected async startInternal(): Promise<void> {
         if (!this.blockProcessor.started) this.logger.error("The BlockProcessor should be started before the BlockchainMachine.");
-        if (!this.actionStore.started) this.logger.error("The ActionStore should be started before the BlockchainMachine.");
+        if (!this.actionStore.started) this.logger.error("The CachedKeyValueStore should be started before the BlockchainMachine.");
```
yahgwai: The message should probably still say actionStore, if that's what we call the property

bigspider: Fixed in f805a6e.

packages/block/src/cachedKeyValueStore.ts
bigspider: Added some docs in f805a6e.

Issue #380, opened 2019-12-10
Long down time can result in missing actions emitted by blockchain machine


yahgwai:
If the machine is down for a long time, when it starts up again it will process all blocks between then and now. Eg 300 blocks. It will call detectChanges only once, using the anchor state of then and of now. Anchor states may pass through intermediary phases between then and now (including items being deleted in the process), which means data and possibly actions are lost. This isnt a problem with the watcher/responder components since they rely on a list of items in aux state, which are only deleted via an action, however this isnt generally true for all components. 

A naive solution would be to call detect changes on all the states between then and now, and run all the aggregated actions. However since the order of action execution cannot be guaranteed we should offer the ability to reduce the action set by removing redundant actions that have been superceded by later ones.

This is a problem for the components in the payment gateway since they do not relay on an external list, but instead purely on the difference between then and now state.

yahgwai:
We also need to make getInitialState and reduce async to enable network calls, and database access.

PR #381, opened 2019-12-11
Detect changes for every block

yahgwai:
We need detectChanges to be called on every block. See #380 . But at the moment we only detect changes/emit when have a new head. New heads can be emitted very far apart and not all the blocks between them may be in the cache. So, if we want this functionality, I think we have 2 choices,

1. Emit heads closer together so that all blocks in between are in the cache
2. Stop detecting changes on new head and instead do it the new block event

We originally had both new head and new block events because we wanted to aggregate changes over large periods of time, since some actions may cancel each other out. However in practice we have to program as if the actions come one at a time as this is what will happen for short gaps between blocks, which is the norm. So bulk detectChanges is just an optimisation, not absolutely necessary. So I've gone for option 2, and we now no longer use the newHead in the blockchain machine.

### DIFF
--- a/packages/block/__tests__/blockchainMachine.test.ts
+++ b/packages/block/__tests__/blockchainMachine.test.ts
@@ -42,10 +42,10 @@ const initialState = {
 };
 
 class ExampleReducer implements StateReducer<ExampleState, IBlockStub> {
-    getInitialState(block: IBlockStub) {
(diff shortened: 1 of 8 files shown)


yahgwai:
Yep, I agree with your comments.

PR #388, opened 2020-01-07
387 publishing

yahgwai:
Simplifies the publishing workflow - still found in publishing.md


docker/dev.Dockerfile
```diff
@@ -7,31 +7,24 @@ WORKDIR /usr/pisa
...
 # we cant use npm run start since it causes problems with graceful exit within docker
 # see https://medium.com/@becintec/building-graceful-node-applications-in-docker-4d2cd4d5d392 for more details
-CMD ["node", "./server/lib/startUp.js"]
-
-
-
-# option1 add the whole context, install lerna, install the dev context
-
-# option2 - add only the server as context, install dev, build, copy to prod image, install prod
+CMD ["node", "./dist/startUp.js"]
```
bigspider: Very minor note, but we could change the name of the fine to index.js now that the server is in its own subpackage (also to stay consistent with any.sender).

PR #389, opened 2020-01-08
Compute initial state upon machine startup

yahgwai:
Allows us to be sure that when the machine has started initial state exists. Especially useful for long running initial state commands.


packages/block/src/blockchainMachine.ts
```diff
@@ -1,27 +1,29 @@
...
+export class BlockchainMachine<TBlock extends IBlockStub> {
     // lock used to make sure that all events are processed in order
     private lock = new Lock();
 
+    constructor(
+        private readonly actionStore: CachedKeyValueStore<ComponentAction>,
+        public readonly blockItemStore: BlockItemStore<TBlock>,
+        private readonly components: Component<AnchorState, TBlock, ComponentAction>[]
+    ) {
+        const duplicateNames = components.map(c => c.name).filter((n, i) => components.map(c => c.name).lastIndexOf(n) !== i);
```
bigspider: This is O(n^2); I guess no big deal for now since we don't have too many components, but in principle it could be possible to have - say - tens of thousands of components; but I'd document the inefficiency with a comment.

packages/block/src/blockchainMachine.ts
```diff
@@ -34,82 +36,125 @@ export class BlockchainMachine<TBlock extends IBlockStub> extends StartStopServi
...
+        if (!this.actionStore.started) logger.error("The actionStore should be started before the BlockchainMachine.");
+        if (!this.blockItemStore.started) logger.error("The BlockItemStore should be started before the BlockchainMachine.");
 
-        this.components.push(component);
-        this.componentNames.add(component.name);
+        try {
+            await this.lock.acquire();
+            // For each component, load and start any action that was stored in the actionStore
+            for (const component of this.components) {
+                // if the current state is not already i n the store we need to compute it and set it there
```
bigspider: extra space in comment

packages/block/src/blockchainMachine.ts
```diff
@@ -34,82 +36,125 @@ export class BlockchainMachine<TBlock extends IBlockStub> extends StartStopServi
...
+     * called on the component reducer.
+     * This function writes to the blockItemStore and MUST be included in a batch to persist beyond memory.
+     * @param block
      */
-    public addComponent(component: Component<AnchorState, TBlock, ComponentAction>): void {
-        if (this.started) throw new ApplicationError("Components must be added before the BlockchainMachine is started.");
-        if (this.componentNames.has(component.name)) throw new ApplicationError(`A Component with the name "${component.name}" was already added.`);
+    public async setInitialState(block: TBlock): Promise<void> {
+        if (!this.actionStore.started) logger.error("The actionStore should be started before the BlockchainMachine.");
+        if (!this.blockItemStore.started) logger.error("The BlockItemStore should be started before the BlockchainMachine.");
```
bigspider: The error messages don't make too much sense now that this is separated from the `BlockchainMachineService`, as the BlockchainMachine is not really "started". Same below in `setStateAndDetectChanges`.

packages/test-utils/src/throwingInstance.ts
```diff
@@ -6,7 +6,9 @@ import * as chai from "chai";
  */
 export function throwingInstance<TMock extends object>(target: TMock) {
     const stubbedMethods: Array<string> = Object.keys(
-        (target as any)["tsMockitoInstance"]["mocker"]["methodStubCollections"]
+        (target as any)["tsMockitoInstance"]["mocker"]
+            ? (target as any)["tsMockitoInstance"]["mocker"]["methodStubCollections"]
+            : (target as any)["mocker"]()["mocker"]["methodStubCollections"]
```
bigspider: I have no idea what's happening here - is it easy to explain in a comment?

packages/block/src/blockchainMachine.ts
yahgwai: updated

packages/block/src/blockchainMachine.ts
yahgwai: fixed

packages/block/src/blockchainMachine.ts
yahgwai: fixed

packages/test-utils/src/throwingInstance.ts
yahgwai: Added a comment

PR #396, opened 2020-02-24
Added blockchain machine end to end tests

### LINKED ISSUE #386: Write integration tests for blockchain machine
```
describe("Blockchain machine integration", () => {
    it("does process new anchor states and detect changes")
    it("does detect changes for all intermediary states")

    it("start, stop, emit blocks, start - does detect all changes")
    it("start, stop, emit blocks greater than cache depth, start - does detect all changes")
    it("start, emit block from x in the future, does detect all changes between now and then")
    it("start, reorg, reorg back to same point, does not detect changes twice")
    it("start, emit block from 10 in the past, emit block now, does detect all changes")
})
```

yahgwai:
closes #386


yahgwai:
All in 386 are done except
```
    it("start, reorg, reorg back to same point, does not detect changes twice")
    it("start, emit block from 10 in the past, emit block now, does detect all changes")
```

We need to use the ganche fork functionality for these - cant just do evm revert - which is a bit more complex so I've moved this to a new ticket.

PR #420, opened 2020-05-15
Logger refactor

bigspider:
- Refactored the Logger, added the "debug" log level.
- Removed the global `logger` instance of Logger, injected logger in all classes instead.
- Add ability to change the log level on a Logger instance.
- Upgrade TypeScript to 3.9.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,6 @@
         "@pisa-research/server": "file:packages/server",
         "@pisa-research/test-utils": "file:packages/test-utils",
         "@pisa-research/utils": "file:packages/utils",
-        "cross-fetch": "^3.0.4",
         "yargs": "^15.1.0"
     },
     "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
         "@pisa-research/server": "file:packages/server",
         "@pisa-research/test-utils": "file:packages/test-utils",
         "@pisa-research/utils": "file:packages/utils",
+        "cross-fetch": "^3.0.4",
         "yargs": "^15.1.0"
     },
     "devDependencies": {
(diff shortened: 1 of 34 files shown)


package.json
```diff
@@ -62,6 +62,7 @@
         "@pisa-research/server": "file:packages/server",
         "@pisa-research/test-utils": "file:packages/test-utils",
         "@pisa-research/utils": "file:packages/utils",
+        "cross-fetch": "^3.0.4",
```
yahgwai, line 65: Why is this added to the root package.json?

packages/utils/src/logger.ts
```diff
@@ -43,42 +22,90 @@ export interface BaseLogObject {
...
+                throw new ArgumentError(`"${levelName}" is not a valid log level name.`);
+        }
+    }
+
+    private readonly children: Logger[] = [];
+
     private constructor(private readonly pino: pino.Logger) {}
 
-    public static getLogger() {
+    public static getLogger(name = "not-set", level: LogLevel = enabled ? "debug" : "silent") {
```
yahgwai, line 60: How will this work with the Config<T> params - since we want to be able to change the log level. Maybe we need to move the Config<T> object into the pisa repo?

bigspider: I'm not sure moving the Config here solves the issue: to change the config level, you need to explicitly set pino.level; instead, the Config level silently changes the value, so it works for settings where we read that value every once in a while.

If we want to use the Config, I guess the options are:
- have a timer to periodically check the logLevel in the Config, and update pino if needed
- do the check every time a function in our Logger is called, before forwarding to pino.

Both seem bad hacks, though. I'd rather put this on a higher level (that is, the place that reads the service config and creates the logger object, sets the level explicitly every time the config is reloaded.

package.json
bigspider: Oops, not sure.
Removed.

packages/utils/src/logger.ts
yahgwai: kk, sounds good

